PR #116, opened 2020-02-18
Fix #87 | Add type to button to prevent it triggering submit

### LINKED ISSUE #87: ListeningChallenge: voice recording is unnnecessarily repeated
Once the challenge is successfully completed, the voice recording is repeated. This is not as intended, since this is completely unnecessary. It's possibly a bug resulting from how and when the sound is played, maybe related to component life-cycle management

Roshanjossey:


kantord:
thanks for the MR! Something seems to be not working with the CI? :thinking: Can you please try rebasing it to master and squashing your MR? I don't know why GitLab CI is not triggered at all

Roshanjossey:
I tried this. It results in the same error. Let me try opening another pull request.

Roshanjossey:
I tried another pull request and it also had the same error

kantord:
I realized what the issue is. Secrets cannot be shared with forks, since then it would be trivial to steal them. (You could just print them out to the console).

Therefore, to make the review app deployment work, you have to set the token for yourself. That's not ideal, but I have no better idea right now :thinking: 
 
Let's try this:

- create a `ZEIT_TOKEN` like this: https://docs.semaphoreci.com/use-cases/zeit-now-continuous-deployment/
- add it as a secret in your fork like this: https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets

PR #133, opened 2020-02-21
Add OptionChallenge

kantord:


src/components/OptionChallenge/index.svelte
```diff
@@ -0,0 +1,75 @@
...
+    resolveChallenge();
+  };
+
+  $: submitChallenge = () => {
+    registerResult(options[selectedOption].correct);
+    submitted = true;
+  };
+
+  onMount(() => {
+    hotkeys.unbind("enter");
```
Roshanjossey: Do you think it makes select options with indices like deck challenge?

PR #164, opened 2020-03-04
Link Read the Docs page in README

kantord:


README.md
```diff
@@ -95,39 +95,8 @@ Testing is done using Jest and Cypress. Other technology in use include: bulma,
 
 ## Development
 
-### Running the project in development mode
+[Development documentation is available on Read the Docs](https://readthedocs.org/dashboard/import/)
```
Roshanjossey: I'm keen to know why this is being moved to an external service as opposed to something like `Contributing.md`.

kantord: hmm well it's technically still just .md files, it's just a fancy viewer for them. It's still this folder: https://github.com/kantord/LibreLingo/tree/master/docs

kantord: there's just too many things to document, like, translation, course editor, frontend, etc. so it's already maybe too big for 1 file.

I'm open to other suggestions tho.

Roshanjossey: Gotcha. Makes sense. It'll be easier to manage the complexity this way as the different parts become more complex.

At the same time, I expected the project to split in to multiple projects as different components mature.

kantord: That will still happen, once they become stable enough so that changes will not have to be coordinated frequently. (For example now you can create a single merge request that implements a UI change and the required backend changes + the documentation all in one go)

PR #274, opened 2020-03-28
Resolves issue #106

kylepollina:
I did not find any images for protest and network


kantord:
Amazing, thanks @kylepollina!

@allcontributors please add  @kylepollina for code

kantord:
The images look perfect! :smile_cat: I only replaced the images about "sand", because I thought that the ones you chose could be confused with desert or dune. The rest were very good choices, thanks again! :smile_cat:

PR #282, opened 2020-03-28
Test case for exportCourse

iasonasma:
Wrote some test cases.


course/tests.py
```diff
@@ -1,3 +1,42 @@
...
-# Create your tests here.
+
+class CommandTests(TestCase):
+    databases = '__all__'
+    fixtures = ["dumps/courseData.json"]
+
+    def test_command(self):
+        out = StringIO()
+        management.call_command('exportcourse', 1, stdout=out, verbosity=3)
+        # print(out.getvalue())
```
kantord: please don't include commented out code

course/tests.py
```diff
@@ -1,3 +1,42 @@
...
+        # self.assertIn("Exporting course Spanish for English speakers course", out.getvalue())
+
+    def test_language_preserved(self):
+        out = StringIO()
+        with translation.override('en'):
+            management.call_command('exportcourse', 2, stdout=out)
+            self.assertEqual(translation.get_language(), 'en', "Same language")
+            # self.assertEqual(translation.get_language(), 'fr', "Not Same language")
```
kantord: same

course/tests.py
```diff
@@ -1,3 +1,42 @@
...
+from django.core import management
+from io import StringIO
+from django.test.utils import extend_sys_path
+from django.utils import translation
 
-# Create your tests here.
+
+class CommandTests(TestCase):
+    databases = '__all__'
+    fixtures = ["dumps/courseData.json"]
```
kantord: I like that idea!

course/tests.py
```diff
@@ -1,3 +1,42 @@
...
+        # self.assertIn("Exporting course Spanish for English speakers course", out.getvalue())
+
+    def test_language_preserved(self):
+        out = StringIO()
+        with translation.override('en'):
+            management.call_command('exportcourse', 2, stdout=out)
+            self.assertEqual(translation.get_language(), 'en', "Same language")
+            # self.assertEqual(translation.get_language(), 'fr', "Not Same language")
+
+    def test_exportcourses(self):
```
kantord: I don't really understand the purpose of this test. What does it test?

course/tests.py
```diff
@@ -1,3 +1,42 @@
...
+        """ An unknown command raises CommandError """
+        with self.assertRaisesMessage(CommandError, "Unknown command: 'exportcourses'"):
+            management.call_command(('exportcourses',))
+
+    def test_discover_commands_in_eggs(self):
```
kantord: What is the purpose of this?

course/tests.py
```diff
@@ -1,3 +1,42 @@
...
+from django.test.utils import extend_sys_path
+from django.utils import translation
 
-# Create your tests here.
+
+class CommandTests(TestCase):
+    databases = '__all__'
+    fixtures = ["dumps/courseData.json"]
+
+    def test_command(self):
```
kantord: I don't see any assertions in this test case :thinking:

iasonasma:
Please check new commit

course/tests.py
```diff
@@ -1,3 +1,29 @@
...
+        sys.stdout = out = StringIO()
+        management.call_command('exportcourse', 1, stdout=out, verbosity=3)
+        self.assertIn("Exporting course Spanish for English speakers course", out.getvalue())
+
+    def test_calling_a_command_without_argument(self):
```
kantord: I don't understand what this is testing? Isn't this logic included in django? We only need to test our own code, not django default functionality

course/tests.py
```diff
@@ -1,3 +1,29 @@
...
+        with self.assertRaisesMessage(CommandError, 'Error: the following arguments are required: course_id'):
+            sys.stdout = out = StringIO()
+            management.call_command('exportcourse', stdout=out)
+
+    def test_exportcourses(self):
+        with self.assertRaisesMessage(CommandError, "Unknown command: 'exportcourses'"):
+            management.call_command(('exportcourses',))
+
+    def test_check_course_does_not_exists(self):
```
kantord: This test case looks good! But maybe use a diffferent id, because 3 might exist in the future. I'm pretty sure something like 345345 would never exist (and if it ever gets to that, we'll have a real text fixture here instead of the actual data anyway)

kantord:
Please move course/tests.py to the same directory the tested stuff are. course/management/commands/test_exportcourse.py or something like that? :thinking:

course/tests.py
iasonasma: Which line number are you talking about? 15 or 17?

course/tests.py
iasonasma: Okay, will add large number.

course/tests.py
iasonasma: In line 15, just checking if this text is in console output or not.

iasonasma: In line 17, we are checking if we pass the argument or not.

iasonasma:
Moving tests.py to another folder doesn't recognize by python manage.py test command.

iasonasma:
Please check another new commit

iasonasma:
Moves tests in tests package with specific file

course/tests/tests_exportcourse.py
```diff
@@ -0,0 +1,29 @@
...
+        management.call_command('exportcourse', 1, stdout=out, verbosity=3)
+        self.assertIn("Exporting course Spanish for English speakers course", out.getvalue())
+
+    def test_calling_a_command_without_argument(self):
+        with self.assertRaisesMessage(CommandError, 'Error: the following arguments are required: course_id'):
+            sys.stdout = out = StringIO()
+            management.call_command('exportcourse', stdout=out)
+
+    def test_exportcourses(self):
+        with self.assertRaisesMessage(CommandError, "Unknown command: 'exportcourses'"):
```
kantord: but `exportcourses` exists

course/tests/tests_exportcourse.py
```diff
@@ -0,0 +1,29 @@
...
+            management.call_command(('exportcourses',))
+
+    def test_check_course_does_not_exists(self):
+        with self.assertRaisesMessage(CommandError, 'Course "645343" does not exist'):
+            sys.stdout = out = StringIO()
```
kantord: do we really need to assign to `sys.stdout` here?

course/tests/tests_exportcourse.py
iasonasma: We have `exportcourse` not `exportcourses`

course/tests/tests_exportcourse.py
iasonasma: `sys.stdout` removed in new commit.

course/tests/tests_exportcourse.py
kantord: since this command actually doesn't exist, we don't need this test case. It's just testing Django's built in functionality. Please remove `test_exportcourses`

course/tests/tests_exportcourse.py
```diff
@@ -0,0 +1,29 @@
...
+    fixtures = ["dumps/courseData.json"]
+
+    def test_command(self):
+        sys.stdout = out = StringIO()
+        management.call_command('exportcourse', 1, stdout=out, verbosity=3)
+        self.assertIn("Exporting course Spanish for English speakers course", out.getvalue())
+
+    def test_calling_a_command_without_argument(self):
+        with self.assertRaisesMessage(CommandError, 'Error: the following arguments are required: course_id'):
+            sys.stdout = out = StringIO()
```
kantord: we still have `sys.stdout` here, I think it's not necessary? :thinking:

course/tests/tests_exportcourse.py
```diff
@@ -0,0 +1,29 @@
...
+            management.call_command('exportcourse', stdout=out)
+
+    def test_exportcourses(self):
+        with self.assertRaisesMessage(CommandError, "Unknown command: 'exportcourses'"):
+            management.call_command(('exportcourses',))
+
+    def test_check_course_does_not_exists(self):
+        with self.assertRaisesMessage(CommandError, 'Course "645343" does not exist'):
+            out = StringIO()
```
kantord: I think this variable can be removed, and instead StringIO() directly passed to `stdout=` on the next line

course/tests/tests_exportcourse.py
```diff
@@ -0,0 +1,29 @@
...
+
+    def test_command(self):
+        sys.stdout = out = StringIO()
+        management.call_command('exportcourse', 1, stdout=out, verbosity=3)
+        self.assertIn("Exporting course Spanish for English speakers course", out.getvalue())
+
+    def test_calling_a_command_without_argument(self):
+        with self.assertRaisesMessage(CommandError, 'Error: the following arguments are required: course_id'):
+            sys.stdout = out = StringIO()
+            management.call_command('exportcourse', stdout=out)
```
kantord: `stdout=StringIO()` should be enough, no?

course/tests/tests_exportcourse.py
```diff
@@ -0,0 +1,29 @@
...
+from django.core import management
+from io import StringIO
+import sys
+
+
+class CommandTests(TestCase):
+    databases = '__all__'
+    fixtures = ["dumps/courseData.json"]
+
+    def test_command(self):
```
kantord: this could be called `test_exportcourse` to be more specific

iasonasma: Okay

course/tests/tests_exportcourse.py
iasonasma: Okay, Removing.

course/tests/tests_exportcourse.py
iasonasma: Okay, Removing.

iasonasma:
Please check new commit

course/tests/tests_exportcourse.py
```diff
@@ -0,0 +1,23 @@
...
+class CommandTests(TestCase):
+    databases = '__all__'
+    fixtures = ["dumps/courseData.json"]
+
+    def test_exportcourse(self):
+        sys.stdout = out = StringIO()
+        management.call_command('exportcourse', 1, stdout=out, verbosity=3)
+        self.assertIn("Exporting course Spanish for English speakers course", out.getvalue())
+
+    def test_calling_a_command_without_argument(self):
```
kantord: actually, this one is also Django built in functionality, so we don't need it.

The other 2 are correct now, so just remove this and I'll merge :smile_cat:

iasonasma: Ah okay

kantord:
thanks @iasonasma!

kantord:
@allcontributors please add @iasonasma for code, test

PR #285, opened 2020-03-29
Exportcourse test

iasonasma:
Added new case


course/tests/tests_exportcourse.py
```diff
@@ -17,3 +17,7 @@ def test_exportcourse(self):
     def test_check_course_does_not_exists(self):
         with self.assertRaisesMessage(CommandError, 'Course "645343" does not exist'):
             management.call_command('exportcourse', 645343, stdout=StringIO(), verbosity=3)
+
+    def test_argument_type(self):
```
kantord: this argument type check is a default Django feature, not a feature of the tested code, hence we don't need to test this, it's already tested at Django probably.

I'll prepare a short MR know to show that type of tests we need more of here

course/tests/tests_exportcourse.py
```diff
@@ -17,3 +19,8 @@ def test_exportcourse(self):
     def test_check_course_does_not_exists(self):
         with self.assertRaisesMessage(CommandError, 'Course "645343" does not exist'):
             management.call_command('exportcourse', 645343, stdout=StringIO(), verbosity=3)
+
+    def test_language_preserved(self):
```
kantord: What does this do?

iasonasma: Stores the current language on enter and restores it on exit and testing if same language is there when exit.

kantord: I don't think this is something we need to test. We have to test our own application code, for example:https://github.com/kantord/LibreLingo/pull/286/files#diff-64fb064dc766bae23b448be12b59a6e0R1

PR #321, opened 2020-04-06
Rebasing old test cases

iasonasma:
Hello.
I've made some new test cases,but also rebased the old one  I had commited (better method of checking).
test_exportcourse -> will see all available courses
test_course_all_skills: In exportcourse.py in line 189 there is a output the skills which the subject loads.We iterate all skills, from courseData.json, and look if everything has been included when course has been exported (means it is  shown on output the line)

test_files_created : in exportcourse.py in line 169-172 it creates the challenges of the subject( the json files from the path src/courses/spanish-from-english/challenges).We take the names of each  module's skill, where it belongs to the subject and we check if every file is created.Also we check if the files of metadata  of course has been created to path : LibreLingo/src/courses/spanish-from-english


course/__init__.py
```diff
@@ -0,0 +1,4 @@
+import pandas as pd
```
kantord, line 1: I don't think we need this dependency in this project right now

course/__init__.py
```diff
@@ -0,0 +1,4 @@
...
+
+test_data = pd.read_json("./dumps/courseData.json");
+test_data = {k: val.groupby('pk')['fields'].apply(list).apply(lambda x: x[0]).to_dict() for k, val in test_data.groupby("model")};
```
kantord, line 4: this piece of code is hard to read, it's unclear what it does.

besides, why is anything here in this file? That smells fishy

course/test_exportcourse.py
```diff
@@ -15,16 +17,61 @@ class CommandTests(TestCase):
 
     def test_exportcourse(self):
         sys.stdout = out = StringIO()
+        course_id = next(iter(test_data['course.course'].keys()))  # Get the first available course
```
kantord, line 20: general note regarding comments of this style ( you have more comments of this sort in your code)

it's unnecessary to put a comment if it can be easily extracted to a function. The benefit of that is that the code is much neater, easier to read.

For example, if you have this line

```python
course_id = next(iter(test_data['course.course'].keys()))  # Get the first available course
```

It can be replaced by creating a new function:

```python
def get_first_available_course():
    return next(iter(test_data['course.course'].keys()))
```

and then it can just be called like this:

```python
course_id = get_first_available_course()
```

It's true that this function is only called in one place, and technically doesn't do much. It's good to minimize the number of lines in your code, but I think what really matters is the number of lines that you actually have to read. So creating a function like that sounds like a good deal to me :smile_cat:

course/test_exportcourse.py
```diff
@@ -15,16 +17,61 @@ class CommandTests(TestCase):
...
+        course_id = next(iter(test_data['course.course'].keys())) # get the first available course
+        course_data = test_data['course.course'][course_id]
+        exportPathString = "./src/courses/{0}-from-{1}".format(course_data['language_name'],
+                                                                course_data['source_language_name'])
+        # check for course metadata file
+        try:
+            if not path.isfile(exportPathString + '/courseData.json'):
+                print('Course\'s metadata file was not created')
+        except:
+            print('Course\'s metadata file path is not created')
```
kantord, line 61: Why are you using print here?

kantord:
Since the tests are creating files, I think the file manipulation should either be mocked, or a fake filesystem should be used

course/test_exportcourse.py
```diff
@@ -53,6 +103,8 @@ def test_returns_correct_value(self, clean_word):
     def test_calls_clean_word_correct_times(self, clean_word):
         generate_chips('foo bar bazz')
         self.assertEqual(clean_word.call_count, 3)
+<<<<<<< HEAD
```
kantord, line 106: oops, that looks like a mistake. This rebase didn't go well

PR #344, opened 2020-04-12
chore: hide loading screen in cypress to increase performance

kantord:


kantord:
closing because it had no real impact on performance

PR #354, opened 2020-04-13
Testcase : opaqueid & audioid

iasonasma:
Hello!

I'm adding some new test ases for "opaqueId" and "audioId" for two concrete values and checking their returns !


kantord:
instead of "not_static" you could write "unique"

kantord:
you have a line that's too long, see CI output: https://kantord.semaphoreci.com/jobs/861ebcc6-58b6-4945-beb7-19cca4062b35

iasonasma:
Hey thanks for the information.

I've done the changes,tests seems to be passed on  CI

kantord:
Thanks! I'm merging this, but if you are looking for a new challenge, you can create another MR with a new test.

It would test the same 2 functions, but actually test one return value.

For example:

```python
def test_audio_id_return_value(self):
        self.assertNotEqual(audioId("1", "foo"), "2df72b91dbc0e1676f9e76afe7c15df0")
```

If you don't know what the correct hash value is, you can simply put whatever else you like. When the test fails, it will show you what value it expected your fake value to be equal to. You can simply copy that value into your unit test.

PR #355, opened 2020-04-13
Adding test case for opaqueid & audioid

iasonasma:
Hey!

I made some new test case for uniqueid and audioid.
It will be checking if the return value is the correct one.


course/test_exportcourse.py
```diff
@@ -40,6 +40,13 @@ def test_argument_type(self):
                 stdout=StringIO(),
                 verbosity=3)
 
+    def test_audio_id_return_value(self):
+        self.assertEqual(audioId("1", "foo"), "36d2a6c2d0e94f671e29b7f0f6223b977e495f08b2d067fde87ef18bd6222ec8")
+
+    def test_opaque_id_return_value(self):
+        learnWord = LearnWord.objects.get(pk=1)
+        self.assertEqual(opaqueId(learnWord), "759ec39eac")
```
kantord, line 48: hmm, I think the parameter to `opaqueId` should rather be a string. It's weird to pass a class like that here

kantord: nvm, I was wrong

PR #358, opened 2020-04-13
Test case for define word

iasonasma:
Hello!

I've written two test cases for define word,one checking if word exists,and if word doesnt exist.

Cheers!


course/test_exportcourse.py
```diff
@@ -59,6 +60,19 @@ def test_opaque_id_unique_with_salt(self):
     def test_audio_id_unique(self):
         self.assertNotEqual(audioId("1", "foo"), audioId("2", "bar"))
 
+    def test_non_existing_word(self):
+        course = Course.objects.get(pk=1)
+        fooObject = {"word": "foo"}
```
kantord, line 65: please try to use a better name here, something like `expectedResult`

course/test_exportcourse.py
```diff
@@ -59,6 +60,19 @@ def test_opaque_id_unique_with_salt(self):
...
+        self.assertEqual(define_word(course, "foo", True), fooObject)
+
+    def test_existing_word(self):
+        course = Course.objects.get(pk=1)
+        validObject = {
+            'word': "dog",
+            'definition': "perro"
+        }
+        self.assertEqual(define_word(course, "dog", True), validObject)
```
kantord, line 74: I'd rather name it "expected_result" instead of "validObject'

course/test_exportcourse.py
```diff
@@ -59,6 +60,19 @@ def test_opaque_id_unique_with_salt(self):
     def test_audio_id_unique(self):
         self.assertNotEqual(audioId("1", "foo"), audioId("2", "bar"))
```
kantord, line 62: I would group these tests in a new `TestCase`, for example:

```
 class DefineWordTest(TestCase):
```

kantord: Also, you will probably need to use this:

```
    databases = '__all__'
```

iasonasma:
Hi.
I've done the changes,also i made class for each method ,so we can have better quality code and to use the test case with the method when called

Cheers!

course/test_exportcourse.py
```diff
@@ -40,9 +41,22 @@ def test_argument_type(self):
                 stdout=StringIO(),
                 verbosity=3)
 
+
+class audioIdTest(TestCase):
```
kantord, line 45: AudioIdTest

course/test_exportcourse.py
```diff
@@ -40,9 +41,22 @@ def test_argument_type(self):
...
+    fixtures = ["dumps/courseData.json"]
+
+    def test_audio_id_unique(self):
+        self.assertNotEqual(audioId("1", "foo"), audioId("2", "bar"))
+
     def test_audio_id_return_value(self):
         self.assertEqual(audioId("1", "foo"), "36d2a6c2d0e94f671e29b7f0f6223b977e495f08b2d067fde87ef18bd6222ec8")
 
+
+class opaqueIdTest(TestCase):
```
kantord, line 56: OpaqueIdTest

course/test_exportcourse.py
```diff
@@ -40,9 +41,22 @@ def test_argument_type(self):
                 stdout=StringIO(),
                 verbosity=3)
 
+
+class audioIdTest(TestCase):
+    databases = '__all__'
```
kantord, line 46: Database and fixtures not needed here, because the tests don't use it. Database makes things slower, so let's skip it when it's not necessary

kantord: (only true for AudioIdTest)

iasonasma: My bad,thought i deleted it

iasonasma:
Hey!
I've done the necessary changes!

Cheers!

PR #432, opened 2020-05-03
Implemented 404

navanchauhan:
Adds` mascot-error` and `mascot-error-noshadow` to the images folder and adds a `mascot-error` component

GitHub Pages requires a special 404.html to be created, therefore I have added a `404.html` to the static folder

There is one minor bug though, I could not add the button to go to the homepage as whenever I added it the build crashed. The link to the home button is working in the 404.html and thus should work with `librelingo.app`


navanchauhan:
This is the 404.html page that will be visible for `librelingo.app` https://libre-lingo-git-fork-navanchauhan-master.kantord.now.sh/404.html

src/components/Mascot-error.svelte
```diff
@@ -0,0 +1,16 @@
+<script>
```
kantord, line 1: I would rather extend the existing `<Mascot />` component to accomodate this, by adding a prop. For example

`export let image`

and it could be either `"jetpack"` or `"error"`.

kantord:
I wonder if it would be possible to automatically generate 404.html instead of having to manually maintain it. I think it would be enough to generate it in the[ deploy step in the CI.](https://github.com/kantord/LibreLingo/blob/master/.semaphore/pipeline_2.yml)

kantord:
Could you please create a gherkin feature-file for the 404 page, one like this one: https://github.com/kantord/LibreLingo/blob/master/cypress/integration/fanfareScreen.feature?

It should be pretty simple, it should just verify the contents of the page and have a [visual regression test](https://github.com/kantord/LibreLingo/blob/master/cypress/integration/landingPage.feature#L8).

navanchauhan:
I am really sorry it is taking me more time than usual. I am currently researching related to nCoV-19 and the workload is too much right now

PR #498, opened 2020-06-01
Update index.md: refine start server step

oscarfrancois:


kantord:
Thanks! Actually, you don't have to `cd` into the folder (and shouldn't, based on how yarn workspaces work).

Instead it's enough to say:

```
yarn workspace web dev
```

which is also aliased to make this shorter version also possibe:

```
yarn web dev
```

oscarfrancois:
OK, I will propose a slightly different pull request.

Issue #341, opened 2020-04-12
ChipsChallenge: Enter key not working with "submit" and "continue"


slesingerm:
**Describe the bug**
In the chip challenge, pressing the enter key does nothing and the "submit" and "continue" buttons have to be operated by a left click.

**To Reproduce**
1. Go to 'Verbs, part 1'
2. Encounter a chip challenge
3. Drag chips to their correct places
4. Press enter
5. See error - no reaction

**Expected behavior**
Hitting the enter should press "continue" and "submit", which happens in other challenges

**Screenshots**
![Screenshot from 2020-04-12 17-11-21](https://user-images.githubusercontent.com/62794315/79072760-20ec5100-7ce3-11ea-8cca-6a5dca8cabf5.png)
![Screenshot from 2020-04-12 17-11-35](https://user-images.githubusercontent.com/62794315/79072799-46795a80-7ce3-11ea-8245-ce9139fd0a55.png)

**Desktop (please complete the following information):**
 - OS: ElementaryOS
 - Browser: Firefox

kantord:
Ok, this probably has to do with the keyboard shortcuts or form field focus in ChipsChallenge.svelte

kantord:
Thanks for reporting!

kantord:
@allcontributors please add @slesingerm for bug

PR #583, opened 2020-08-04
fix: fix enter key in ChipsChallenge

dhelmr:
This should fix https://github.com/kantord/LibreLingo/issues/341

It basically adds the same section for binding the enter key as in other files, like `ShortInputChallenge.svelte`. Both continuing and submitting now work for me (Firefox 79.0, Arch Linux).

Sorry if I've oversaw something and `onMount` is not called there for a reason. Also sorry if I violated some code guidelines - svelte is new to me. Making a pull request on github as well, but this project looks pretty cool, so I thought I should contribute somehow :)

### DIFF
--- a/workspaces/web/src/components/ChipsChallenge.svelte
+++ b/workspaces/web/src/components/ChipsChallenge.svelte
@@ -1,5 +1,7 @@
 <script>
+  import { onMount } from "svelte"
   import Sortable from "sortablejs"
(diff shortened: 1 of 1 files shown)


kantord:
Perfect, thank you @dhelmr !

kantord:
@allcontributors please add @dhelmr for code

PR #659, opened 2020-08-29
Reduce length of skill exercises by requiring multiple practices to complete a skill

kantord:


kantord:
It's still WIP, but since you've recently contributed @kivutar @doamatto @LucasCTN (or anyone else interested), would you want to review?

I think PR issue is a nice example of a somewhat complex change in LibreLingo, so it should show a couple of things about how to work with the codebase!

LucasCTN:
> It's still WIP, but since you've recently contributed @kivutar @doamatto @LucasCTN (or anyone else interested), would you want to review?
> 
> I think PR issue is a nice example of a somewhat complex change in LibreLingo, so it should show a couple of things about how to work with the codebase!

I want to take a closer look on it tomorrow, although it seems like I'm very inexperienced in this project to properly review it 😬

slesingerm:
I ran through a course and I had no problems with this, I think the progress bar is a really nice touch and it gave me the feeling I was progressing. It also definitely feels like a more pleasant learning experience doing the course in several "levels" rather than doing one loooong course. 

One thing I did find is that I often had the same sentence in succession in different exercises - "Los patos son médicos" would come as a multiple choice, a writing, a chips and a listening right after each other. Would there be a way to break up these "clusters"?

I was also wondering to myself if adding a "percentage" to the progress bar would be a good idea, but at least for me personally the visual representation is more than enough and I feel like adding a number would just break up the cohesion of the card.

kantord:
> I ran through a course and I had no problems with this

Thank you, @slesingerm!

>  Would there be a way to break up these "clusters"?

It would certainly be possible to break them up into clusters, although that could make things a little bit more difficult to orchestrate. That would probably mean that we would need to keep track in a way of which clusters have been recently practiced, or cycle through them in some way, basically implementing the clusters using [modular arithmetic](https://en.wikipedia.org/wiki/Modular_arithmetic).

I have a couple of concerns with the modular arithmetic version though:

- The practices would be deterministic based on the saved state. In practice this means if you board a plane with a tablet and a phone, and practice on both devices, you will get the *exact* same practice sessions on each device. With the current implementation you get a random hand each time.
- We would also need to orchestrate this in the course editor somehow, so that changes in the course content don't randomize the clusters. (Since ideally in the future we'd like to see frequent changes to the course content)

Those concerns could be mitigated, but that would highly increase complexity, so I'm not sure if it's worth it, given the very limited amount of work hours happening on this project. Either way, unless this repetition issue is a complete deal breaker for this pull request, it should perhaps be de-scoped :smile_cat:

Other than clustering, I can imagine another way this could go: using spaced repetition. Right now, spaced repetition is applied at the skill level, so it doesn't look at each word or phrase. If we applied spaced repetition to each challenge, things that have gone the longest time without practice would always take precedence, therefore all unnecessary repetition would be eliminated.

This could be wonderful, because it would also let us consider errors committed with each challenge: words or phrases with high error rates would get practiced somewhat more frequently. 

The big issue I see with that is that it would increase the amount of data stored by at least an order of magnitude, meaning the same server would be able to host 10 times less users, and the chance that the app would be affected by storage limitation on the client side would also increase, so this is a change I would only consider after some deeper investigation.

What can maybe done as a quick fix is reducing the number of levels per skill. Once the skill is fully learned, most users will probably stop practicing them until they get stale, so the change they get fatigued by the repetition is lowered.

What do you think?

kantord:
@allcontributors please add @slesingerm for review

PR #755, opened 2020-10-04
Add liblili2json

kantord:


LucasCTN:
I liked, looks way more readable format and looks great for this application.
The challenges would be described on the hello.yaml file in this example, right?

garrison:
I might have missed it, but do you have any plan for how to reference images in the yaml course materials?

The Typo CI checks are quite annoying on github's "files changed" tab.

kantord:
> The Typo CI checks are quite annoying on github's "files changed" tab.

Yes, I'm considering getting rid of those

kantord:
> I might have missed it, but do you have any plan for how to reference images in the yaml course materials?

Currently the images are referenced [by name](https://github.com/kantord/LibreLingo/tree/master/workspaces/web/static/images). This is just because that's how it's implemented in the current course editor.

Once this thing is working, as a next step I plan to implement a feature where you could just put an [Unsplash](https://unsplash.com/) URL, and the image would be automatically fetched. (Also ideally added to some credits/attributions page)

kantord:
>  The challenges would be described on the hello.yaml file in this example, right?

Yes, indirectly. Instead of containing specific challenges, it's containing words and sentences. From those, there will be a script generating descriptions of the challenges.

The reason is that it would be very redundant to describe challenges, because from the same word, a variety of different challenges can be created for example.

kantord:
I'll merge this because it it's already growing huge! Even thought it still hasn't achieved what I originally set out to achieve here, it doesn't have any breaking changes and increases test coverage, so there's enough value to merge.

It's about 70% the way to being able to export from yaml based courses. I'll try to implement the remaining part in a couple small MRs

Issue #753, opened 2020-10-04
Cypress: disable sounds effects


kantord:
Cypress tests interact with parts of the application where sounds effects are made. This is distracting when working with the tests locally. (i. e. `yarn web cy:open` or `yarn web cy:run`)

`howler` is used project-wide to implement sound effects, here's **one example**:

https://github.com/kantord/LibreLingo/blob/c89466d7560eb3338b95d4c2cdb7fe02d1b51af4/workspaces/web/src/media/voice.js#L1-L9

The task is to make sure that the sound effects are not played when running from Cypress. The easiest way to detect that a code is running from Cypress is using `window.isCypress`.

Ideally, this solution would not involve having to check in each place where sound is played. Instead, this logic should be implemented in one place, and simply re-used in all the other places. This can be done, for example, by subclassing `howler`, or whatever solution that seems fit!

DavidLuterancik:
I am looking for my first contribution and would like to try implementing a solution for this issue.

luthrap:
I would like to participate as well.

kantord:
@DavidLuterancik awesome, go for it!! If you have any questions, feel free to ask

@luthrap let's give this one to @DavidLuterancik. Of course if you agree to, you can cooperate/do pair programming etc :wink: Otherwise, there's some more[ good first issues](https://github.com/kantord/LibreLingo/issues?q=is%3Aopen+is%3Aissue+label%3AHacktoberfest+no%3Aassignee) that you can check out:/

spankyed:
For the love of god please mute sounds on headless tests.

imohib:
Any progress?

alexqem:
Also looking for solution. Working with apps with sound effects

ucatunit:
I struggled with this as well - app I work with has forced autoplay whenever video player appears. No matter if I was running the tests in headless or through cypress dashboard. I am not familiar with howler at all, as well as maybe I am stating the obvious here, though you may pass additional launchOptions on `before:browser:launch` event in `cypress.config.js`:

```
on(‘before:browser:launch’, (browser = {}, launchOptions) => {
                if (browser.family === ‘chromium’) {
                    launchOptions.args.push(‘--mute-audio’)
                }

                return launchOptions
            })
```

the following piece of code mutes the audio completely.

list of flags: https://github.com/GoogleChrome/chrome-launcher/blob/main/docs/chrome-flags-for-tools.md

This is a bit of overkill, although worked in my case.

[...]

PR #756, opened 2020-10-04
feat(sounds): disabling the sounds when running cypress

DavidLuterancik:
This PR resolves the issue #753 

Disabled the sounds when running cypress tests with additional refactoring to avoid further redundancy on sound files played using Howl.

Please let me know if anything is wrong :)

### DIFF
--- a/workspaces/web/src/media/sound.js
+++ b/workspaces/web/src/media/sound.js
@@ -1,15 +1,17 @@
 import { Howl } from "howler"
 
 const sound = {
-    correct: new Howl({
-        src: ["sound/correct.mp3"],
-    }),
-    wrong: new Howl({
-        src: ["sound/wrong.mp3"],
-    }),
-    fanfare: new Howl({
-        src: ["sound/fanfare.mp3"],
-    }),
+    correct: () => playAudio("sound","correct"),
+    wrong: () => playAudio("sound","wrong"),
+    fanfare: () => playAudio("sound","fanfare")
+}
+
+export const playAudio = (type,filename) => {
+    if(!window.Cypress){
+        new Howl({
+            src: [`${type}/${filename}.mp3`],
+        }).play()
+    }
 }
 
 export default sound
--- a/workspaces/web/src/media/sound.js
+++ b/workspaces/web/src/media/sound.js
@@ -1,9 +1,15 @@
 import { Howl } from "howler"
 
 const sound = {
-    correct: () => playAudio("sound","correct"),
-    wrong: () => playAudio("sound","wrong"),
-    fanfare: () => playAudio("sound","fanfare")
+    correct: new Howl({
+        src: ["sound/correct.mp3"],
+    }),
+    wrong: new Howl({
+        src: ["sound/wrong.mp3"],
+    }),
+    fanfare: new Howl({
+        src: ["sound/fanfare.mp3"],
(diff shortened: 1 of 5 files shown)


workspaces/web/src/media/sound.js
```diff
@@ -1,15 +1,17 @@
...
-export default sound
+export const playAudio = (type,filename) => {
+    if(!window.Cypress){
+        new Howl({
+            src: [`${type}/${filename}.mp3`],
+        }).play()
+    }
+}
+
+export default sound
```
kantord, line 17: Can you pls add a newline at the end of the file?

workspaces/web/src/media/sound.js
```diff
@@ -1,15 +1,17 @@
...
-    correct: new Howl({
-        src: ["sound/correct.mp3"],
-    }),
-    wrong: new Howl({
-        src: ["sound/wrong.mp3"],
-    }),
-    fanfare: new Howl({
-        src: ["sound/fanfare.mp3"],
-    }),
+    correct: () => playAudio("sound","correct"),
```
kantord, line 4: One danger that I see here is that this way the browser will only start downloading the sound file once it should be actually played. That could lead to some extra latency when it's first played, especially on slow networks.

Before this the download of the sound file started (afaik) when the module was imported (i. e. when `new Howl` was called

kantord:
Hi, @DavidLuterancik, thank you!!

I think it's almost perfect, just please resolve the 2 comments that I pointed out

DavidLuterancik:
After testing the latency with simulated slower network, I've reverted to the previous solution with sounds. As you stated, the sounds are downloaded at the module import.

Also, added the new line at the end of the file.

workspaces/web/src/media/sound.js
```diff
@@ -12,4 +12,12 @@ const sound = {
     }),
 }
 
+export const playAudio = (type,filename) => {
+    if(!window.Cypress){
```
kantord, line 16: ok, this is perfect, because this never had a preloading system before anyway. Perhaps, we can create a task for building the preloading system, but it's not relevant here! :+1:

kantord:
Based on the code now it seems to me though, that the "sounds effects" are not disabled at all, only the voices :thinking:  Look for other usages of howler.

DavidLuterancik:
I've tried to find a different solution, this seems to be working fine. I am not quite happy with the way it achieves it. Should I find a way to stop/mute only certain sounds or is this enough for running tests.

image

kantord:
> I've tried to find a different solution, this seems to be working fine. I am not quite happy with the way it achieves it. Should I find a way to stop/mute only certain sounds or is this enough for running tests.
> 
> image

I think this is a good way!

Another way would be to wrap howler sth like this:

```
const loadSound = ({ src }) => {
  if (window.isCypress) {
     return { play: () => null }
  }
 
  return new Howler({ "src": [src] })
}

this way it will be possible to `.play()` it every time, but in cypress it will do nothing.
```

DavidLuterancik:
Couldn't put my finger on it with the wrap :/ hope this will do for now

kantord:
Yes! Looks great to me! :100:  Thank you!

kantord:
@allcontributors please add @DavidLuterancik for code

Issue #722, opened 2020-09-27
Set up TypeScript


kantord:
## What is required
- set up typescript support for all packages under `workspaces/`
- existing JavaScript/Svelte code should work, but adding type annotations optionally should also work

## What is not required
- No need to add type annotation to the entire project, just prove that it works 

## Almost good first issue, BUT
This task doesn't require you to know the project structure, but it might be challenging if you have no previous experience in setting up TypeScript in svelte projects.

**caveat is that we use webpack, not rollup. Likely many code examples will be using rollup**

**caveat2 is that we also have Cypress**

HelioStrike:
@kantord I don't have experience working with svelte projects, but I'd like to give this a peek.

kantord:
Sure, @HelioStrike sure, go for it!

kantord:
this is done, we just forgot to close the issue

PR #757, opened 2020-10-05
feat: set up typescript

HelioStrike:
#722 

## Changes

- Added typescript support to `.svelte` files using `svelte-preprocess` in `webpack.config.js`.
- (Forgetting about the "What is not required" section of the original issue) Changed added `lang="typescript"` and  `type="text/scss"` to `<script>` and `<style>` tags in the `.svelte` files.
- Added `ts-loader` to webpack so that `.js` files can be converted to `.ts`.
- Added `tsconfig.json` (copied from [here](https://github.com/tsconfig/bases/blob/master/bases/svelte.json)).

## Testing

- With the changes made, nothing breaks right now. I also made sure that typescript actually works by pasting a snippet from [here](https://www.typescriptlang.org/docs/handbook/generics.html) into the `<script>` tags. I also converted a `.js` file into `.ts`, pasted the snippet, and ran `yarn web dev`. Both cases returned no errors.

## Issues

- Need to look at `tsconfig.json` and the "TO-DO"s (ctrl+f) in `webpack.config,js`, and update them if necessary.
- [This line](https://github.com/kantord/LibreLingo/blob/master/workspaces/answer-corrector/index.js#L60) returned an error because the function doesn't contain `alwaysSuggest` as an argument. This code works when its `.js` but breaks in `.ts`. I think there might be some compiler option to fix this.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,0 +1,27 @@
+{
+    "$schema": "https://json.schemastore.org/tsconfig",
+    "display": "Svelte",
+  
+    "compilerOptions": {
+        "moduleResolution": "node",
+        "target": "es2017",
+        /** 
+        Svelte Preprocess cannot figure out whether you have a value or a type, so tell TypeScript
+        to enforce using `import type` instead of `import` for Types.
+        */
+        "importsNotUsedAsValues": "error",
+        "isolatedModules": true,
+        /**
+        To have warnings/errors of the Svelte compiler at the correct position,
+        enable source maps by default.
+        */
+        "sourceMap": true,
+        /** Requests the runtime types from the svelte modules by default. Needed for TS files or else you get errors. */
+        "types": ["svelte"],
+
+        "strict": false,
+        "esModuleInterop": true,
+        "skipLibCheck": true,
+        "forceConsistentCasingInFileNames": true
+    }
+}
--- a/package.json
+++ b/package.json
@@ -30,9 +30,12 @@
   },
   "devDependencies": {
     "eslint": "6.8.0",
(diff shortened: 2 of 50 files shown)


kantord:
Thanks! This is amazing!!! I'll need to do a proper review, which I'll do asap (probably tomorrow evening)

> This line returned an error because the function doesn't contain alwaysSuggest as an argument. This code works when its .js but breaks in .ts. I think there might be some compiler option to fix this.

Or rather we can fix that code. Once I'm done with the review, I'll merge this into a temporary branch, so that the CI pipeline and visual regression testing, etc. will be executed. That way we'll see if there's any errors and we can fix them.

tsconfig.json
```diff
@@ -0,0 +1,27 @@
+{
+    "$schema": "https://json.schemastore.org/tsconfig",
+    "display": "Svelte",
+  
+    "compilerOptions": {
+        "moduleResolution": "node",
+        "target": "es2017",
+        /** 
+        Svelte Preprocess cannot figure out whether you have a value or a type, so tell TypeScript
```
kantord, line 9: thanks this comment, it's very useful! :100:

kantord:
> (Forgetting about the "What is not required" section of the original issue) Changed added lang="typescript" and type="text/scss" to <script> and <style> tags in the .svelte files.

that's great!

kantord:
I tried it out here: https://github.com/kantord/LibreLingo/pull/764 and it seems to work perfectly!!!!

thank you very much!!!! :100:

kantord:
Amazing job!

@allcontributors please add @HelioStrike for code

PR #810, opened 2020-10-16
Update LICENSE

anuj840:


kantord:
Hi @anuj840 , what is this? Why did you create this change? What difference does it make?

Issue #806, opened 2020-10-15
Add sponsor button to the page


kantord:
**Is your feature request related to a problem? Please describe.**
The sponsoring information is only present on the GitHub page for now. It would be better if it was also linked in the app itself

**Describe the solution you'd like**
Please add a sponsor button on the web page, next to the github stars button on the navbar.

The sponsor button should lead to this link: https://github.com/sponsors/kantord

This is where you'll have to add the button component: https://github.com/kantord/LibreLingo/blob/3ab9497db7121f38cea88d6aac7423751c19a804/workspaces/web/src/components/NavBar.svelte#L29

Create a simple button component that work similarly to the github button

gkydev:
Hi can you assign this to me

kantord:
Sure @gkydev !

gkydev:
Thank you @kantord ! 

It supposed to look like this right ? 

![image](https://user-images.githubusercontent.com/26537245/96185593-cdcdb200-0f42-11eb-8e63-d887554f1c7c.png)

kantord:
@gkydev yes, that looks perfect!

PR #811, opened 2020-10-16
feat(navbar): added sponsor button

gkydev:
Adds sponsor button 
#806

### DIFF
--- a/workspaces/web/src/components/NavBar.svelte
+++ b/workspaces/web/src/components/NavBar.svelte
@@ -1,6 +1,7 @@
 <script lang="typescript">
   import settings from "../settings"
   import authStore from "../auth"
+  import SponsorButton from "./SponsorButton"
   import GitHubButton from "./GitHubButton"
   import Icon from "lluis/Icon.svelte"
   import Button from "lluis/Button"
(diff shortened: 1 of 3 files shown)


kantord:
hi @gkydev thank you for the contribution!

workspaces/web/src/components/NavBar.svelte
```diff
@@ -26,7 +26,14 @@
   <div class="navbar-end">
     <div class="navbar-item">
       <div class="buttons">
-        <GitHubButton />
+		<Button href="http://github.com/sponsors/kantord" size="small" outlined inverted info>
+		<svg viewBox="0 0 16 16" version="1.1" width="16" height="16" aria-hidden="true" style="fill: #ea4aaa!important;margin-right: 5px">
```
kantord, line 30: Please don't inline this svg. Instead, use an icon from font awesome.

We already have an icon component for this: https://github.com/kantord/LibreLingo/blob/master/workspaces/lluis/Icon.svelte

Here's an example of the icon component being used:
https://github.com/kantord/LibreLingo/blob/ccfaaccf958055069659e19654ad10bdc2002f6e/workspaces/web/src/components/GitHubButton.svelte#L50

If the icon component doesn't currently support the specific icon you need, you can add it here:
https://github.com/kantord/LibreLingo/blob/master/workspaces/web/src/client.js

You can browse the set of the available icons here: https://fontawesome.com/icons/search

If you can't find the same icon that you put here, you can choose whatever else that fits the purpose in your judgement!

workspaces/web/src/components/NavBar.svelte
```diff
@@ -26,7 +26,14 @@
   <div class="navbar-end">
     <div class="navbar-item">
       <div class="buttons">
-        <GitHubButton />
+		<Button href="http://github.com/sponsors/kantord" size="small" outlined inverted info>
```
kantord, line 29: Instead of having this button right here, please move it to a new file, similarly how GitHubButton is implemented

kantord (changes requested):
2 comments, otherwise looks good to me :-)

workspaces/web/src/components/NavBar.svelte
gkydev: Should be ok right now

kantord (approved):

gkydev:
Thank you for this learning oppurtunity @kantord

Issue #181, opened 2020-03-12
CI: add size-limit


kantord:

douglasferlini:
Hi! I would like to start helping.
Could you assign me in this issue?

kantord:
Sure! Thank you!

kantord:
It might be challenging since it won't be super clear what to measure and how, since it's basically a static site

kantord:
So kinda coming up with an idea is part of the task

douglasferlini:
Thanks! I'll be working on it.

douglasferlini:
I've searched on docs and on source code and still not figured out what you mean  by "CI:".
Could you please clarify what it means?

kantord:
@douglasferlini sorry for the late reply!

Continous Integration, i. e. adding it to the semaphore configuration: 
https://github.com/kantord/LibreLingo/tree/master/.semaphore

douglasferlini:
I am trying to test your Semaphore configuration in my on fork, but I've got an error for some missing secrets.
Can you provide these secrets or help me customize mine?

kantord:
Hi @douglasferlini, I put up some documentation regarding this:

https://librelingo.readthedocs.io/en/latest/#setting-up-semaphore-ci-in-a-clone

douglasferlini:
Thanks. I've managed to execute the semaphore workflow in my forked repository.
I will review the sources to identify where do a size limit apply.
If you have any hints on where you would like this limits, please tell me.

kantord:
I would probably do size limits somehow on binaries. As I said it's probably not trivial to do, but the point is to avoid too much data traffic slowing down the website, therefore what we want to measure would be actual bundles/binaries. For example initial bundles.

kantord:
Maybe Sapper has some sort of relevant documentation for this, or maybe you can get help at their GitHub page: https://github.com/sveltejs/sapper

douglasferlini:
Ok, I will start checking on Sapper.

Thanks

douglasferlini:
After studying svelte and sapper a little bit an initial solution could be configure webpack to limit entry points's size and assets's size. 
What do you think?

kantord:
Hi @douglasferlini, sorry for the late reply. If I ever ignore an issue for this long a time, feel free to ping me please :-) 

yes I think limiting the webpack entry size could be a good start indeed!

douglasferlini:
[...]

PR #857, opened 2020-10-25
Issue: CI: add size-limit #181

douglasferlini:
Pull request for issue https://github.com/kantord/LibreLingo/issues/181


workspaces/web/webpack.config.js
```diff
@@ -144,13 +144,20 @@ module.exports = {
         },
         mode: process.env.NODE_ENV,
         performance: {
-            hints: false, // it doesn't matter if server.js is large
```
kantord, line 147: Well, it really kinda doesn't matter if server.js is large, at least not to the same degree that it matters that client.js is large.

I think you meant to introduce a limit on client.js, not server.js.

douglasferlini: I've changed the webpack config to limit only the client size.  Please check if it is what you was expecting.

kantord: yes, it does make sense. How do the set limits compare to the actual numbers observed?

kantord:
Thanks @douglasferlini!

@allcontributors please add @douglasferlini for code

Issue #787, opened 2020-10-11
Update to webpack 5


kantord:
The automatic upgrade failed: https://github.com/kantord/LibreLingo/pull/786

The task is to fix the issues and make the upgrade pass.

This should be a good first issue as it will not require deep knowledge of the project, but keep in mind that there's no guarantees that updating webpack will immediately work, because it's a fresh release.

Official blog on version 5: https://webpack.js.org/blog/2020-10-10-webpack-5-release/

chernykhal:
Hello! Can I try do it?
I'm new at working with webpack and have never worked with svelte projects at all, but I would like to give it a try. I think it will be interesting :smile:

kantord:
Sure, feel free to give it a try!

kantord:
Hi @chernykhal, how is it going? do you need some help?

chernykhal:
@kantord Hey, i'm in progress, but it really looks harder, than i thought
I'm just update webpack package to v5 and it instantly doesn't work
<details>
  <summary>Console output</summary>
`
alexander@alexander:~/sites/forks/LibreLingo/workspaces/web$ yarn dev
yarn run v1.22.5
warning package.json: No license field
$ sapper dev --no-hot
✗ server
resolve 'internal' in '/home/alexander/sites/forks/LibreLingo/workspaces/web/node_modules/svelte'
  using description file: /home/alexander/sites/forks/LibreLingo/workspaces/web/node_modules/svelte/package.json (relative path: .)
    using description file: /home/alexander/sites/forks/LibreLingo/workspaces/web/node_modules/svelte/package.json (relative path: ./internal)
      /home/alexander/sites/forks/LibreLingo/workspaces/web/node_modules/svelte/internal is not a file
(node:6782) [DEP_WEBPACK_TEMPLATE_PATH_PLUGIN_REPLACE_PATH_VARIABLES_HASH] DeprecationWarning: [hash] is now [fullhash] (also consider using [chunkhash] or [contenthash], see documentation for details)
(node:6782) [DEP_WEBPACK_COMPILATION_ASSETS] DeprecationWarning: Compilation.assets will be frozen in future, all modifications are deprecated.
BREAKING CHANGE: No more changes should happen to Compilation.assets after sealing the Compilation.
        Do changes to assets earlier, e. g. in Compilation.hooks.processAssets.
        Make sure to select an appropriate stage from Compilation.PROCESS_ASSETS_STAGE_*.
✗ client
[...]

PR #860, opened 2020-10-26
Issue: Update to webpack 5 #787

chernykhal:
Sorry for waiting :smile:


workspaces/web/webpack.config.js
```diff
@@ -22,6 +22,12 @@ module.exports = {
         resolve: { alias, extensions, mainFields },
         module: {
             rules: [
+                {
+                    test: /\.m?js/,
```
kantord, line 26: why are we adding this rule? Do you have some explanation?

kantord: >  Invalid configuration object. Webpack has been initialised using a configuration object that does not match the API schema.00:01
 - configuration.module.rules[0].resolve has an unknown property 'fullySpecified'. These properties are valid:

This tells be that something is wrong here :thinking:

kantord:
Please make sure that the CI pipeline doesn't fail: https://kantord.semaphoreci.com/jobs/3c864ab8-2ae8-40dd-90ab-491c590035db 

Apparently, sapper build doesn't work after these changes

chernykhal:
I guess, this rule necessary for correctly work with webpack 5 and svelte. By default, you need to specify extension of imported module. When you remove this rule, webpack says that module not found
This is purely a webpack 5 issue nothing to do with Svelte
They have enabled a force fully specified rule to true by default
not sure why

chernykhal:
@kantord Okey, sapper build step is complete, but tests not complete, hmm... :thinking:

chernykhal:
image
That's really bad. I don't know what a problem with vercel's deployment

kantord:
Well, I guess `process`perhaps doesn't exist anymore in webpack 5? :thinking:

chernykhal:
> Well, I guess `process`perhaps doesn't exist anymore in webpack 5?

Yeah, but how does this even affect?

kantord:
> > Well, I guess `process`perhaps doesn't exist anymore in webpack 5?
> 
> Yeah, but how does this even affect?

as you can tell, yes. It's used widely in LibreLingo, which is why the app is broken and the ci check is failing

kantord:
There's also a conflict in package.json now, I guess it's because in the meantime someone else, or the bot, updated a dependency

chernykhal:
I apologize in advance for my illiteracy (technical and grammatical). The documentation states that [the process is no longer supported](https://webpack.js.org/migrate/5/#test-webpack-5-compatibility). But I don't see these options in the config.
Could it be related with constants? 
For example:
`mode: process.env.NODE_ENV,`
`...(process.env.ANALYZE === "true" ? [new BundleAnalyzerPlugin()] : []),`
There is also information about this [(Regarding Runtime Errors)](https://webpack.js.org/migrate/5/#run-a-single-build-and-follow-advises).

kantord:
> I apologize in advance for my illiteracy (technical and grammatical). The documentation states that [the process is no longer supported](https://webpack.js.org/migrate/5/#test-webpack-5-compatibility). But I don't see these options in the config.
> Could it be related with constants?
> For example:
> `mode: process.env.NODE_ENV,`
> `...(process.env.ANALYZE === "true" ? [new BundleAnalyzerPlugin()] : []),`
> There is also information about this [(Regarding Runtime Errors)](https://webpack.js.org/migrate/5/#run-a-single-build-and-follow-advises).

Yeah, it sounds like this is confusing. I might be wrong but I think it just means that it was enabled by default.

Not sure what you mean "Could it be related with constants?", do you mean replacing it with an environmental variable? 

Here's what the webpack 5 docs has to say about it:


2020-10-31-064731_764x557_scrot

One way is to add the polyfill manually, which looks very simple: https://medium.com/@sanchit3b/how-to-polyfill-node-core-modules-in-webpack-5-905c1f5504a0 but apparently the webpack 5 docs doesn't recommend it.

Going with the ENV variable idea should be pretty simple, because we have 2 separate configs for the server, so you could just define, say, the `IS_SSR` variable as true in the server config, and false in the client config. 

I believe this plugin should let your define environment variables in webpack: https://webpack.js.org/plugins/environment-plugin/#usage-with-default-values

chernykhal:
@kantord Hello, sorry for waiting, I was very busy. So, I tried defining variables like you said, but it still doesn't work and I can't figure out how defining the IS_SSR variable will help. Once again, sorry to wait, but now I am free and can do this, although I do not understand some things.

kantord:
> @kantord Hello, sorry for waiting, I was very busy. So, I tried defining variables like you said, but it still doesn't work and I can't figure out how defining the IS_SSR variable will help. Once again, sorry to wait, but now I am free and can do this, although I do not understand some things.

Basically there are 2 config profiles inside the webpack config. Here's the client config: https://github.com/kantord/LibreLingo/blob/master/workspaces/web/webpack.config.js#L19

Here's the server config: https://github.com/kantord/LibreLingo/blob/master/workspaces/web/webpack.config.js#L110

You can simple set the env var to true for server and false for client. Then that variable can be used in the code

chernykhal:
> > @kantord Hello, sorry for waiting, I was very busy. So, I tried defining variables like you said, but it still doesn't work and I can't figure out how defining the IS_SSR variable will help. Once again, sorry to wait, but now I am free and can do this, although I do not understand some things.
> 
> Basically there are 2 config profiles inside the webpack config. Here's the client config: https://github.com/kantord/LibreLingo/blob/master/workspaces/web/webpack.config.js#L19
> 
> Here's the server config: https://github.com/kantord/LibreLingo/blob/master/workspaces/web/webpack.config.js#L110
> 
> You can simple set the env var to true for server and false for client. Then that variable can be used in the code

Like this?

chernykhal:
> You can simple set the env var to true for server and false for client. Then that variable can be used in the code
Like this?
[Client](https://github.com/chernykhal/LibreLingo/blob/697a23d777f966c8a2d5a00f97dc9c599398abac/workspaces/web/webpack.config.js#L113)
[Server](https://github.com/chernykhal/LibreLingo/blob/697a23d777f966c8a2d5a00f97dc9c599398abac/workspaces/web/webpack.config.js#L163)

chernykhal:
@kantord 
I defined the variable IS_SSR, in the client it is false, and in the server it is true, now this definition is needed to use environment variables in the code?
For example
```javascript
if (IS_SSR === true)

PR #871, opened 2020-10-30
feat: implement tag svelte component

### LINKED ISSUE #850: Turn sass/elements/tag.sass into a component
This file is currently imported in [mystyles.scss](https://github.com/kantord/LibreLingo/blob/master/workspaces/web/src/mystyles.scss). It corresponds to a [bulma feature](https://bulma.io/documentation/).

This is all and well, but long term we want to avoid using bulma, and instead implement the styles within each component. The advantage of that is that we don't have to put unused CSS on pages, increasing our performance.

For now, this task consists of

- Creating a new [lluis component using Svelte](https://github.com/kantord/LibreLingo/tree/master/workspaces/lluis) that implements the same feature this sass file does
- Replace all usages of the classes found in this sass file with the component you just implemented.

**You don't need to re-implement the styles**. For now, it's perfectly fine if you just import the same .sass file, but only in the newly creating component. This already helps to achieve this optimization.

If you run into a tricky situation, don't be afraid to ask for help!

This is a very simple example of a component you might have to create:

https://github.com/kantord/LibreLingo/blob/9a84e20e6a14b0de6d31037039fcc5990dc4ed2f/workspaces/lluis/Box.svelte#L1-L3

As you can see, the main idea is implementing a simple component that adds a class.

If the usages involving this component are a little bit more tricky, you might have to introduce some props, as witnessed in this component: https://github.com/kantord/LibreLingo/blob/master/workspaces/lluis/Column.svelte

**No need to implement any component that isn't actually used.**

kenanchristian:
Implemented Tag Component (Fixes #850)

### DIFF
--- a/workspaces/lluis/Tag.svelte
+++ b/workspaces/lluis/Tag.svelte
@@ -1,5 +1,4 @@
 <script lang="typescript">
-  export let isContainer = false
   export let color = null
   export let theme = null
   export let size = null
@@ -8,30 +7,15 @@
   export let style = ""
 </script>
 
-{#if isContainer}
-  <div class="tags {`${ isContainer && size ? `are-${size}` : ''}`}" {style}>
-    <slot />
-  </div>
-{:else}
-  <span {style} class="tag {`tag ${size ? `is-${size}`: ''} ${theme ? `is-${theme}`: ''} ${color ? `is-${color}`: ''}`}"
-  class:is-rounded={isRounded}
-  class:is-delete={isDelete}
-  >
-    {#if !isDelete}
-      <slot/>
-    {/if}
-  </span>
-{/if}
+<span {style} class="tag {`tag ${size ? `is-${size}`: ''} ${theme ? `is-${theme}`: ''} ${color ? `is-${color}`: ''}`}"
+class:is-rounded={isRounded}
+class:is-delete={isDelete}
+>
+  {#if !isDelete}
+    <slot/>
+  {/if}
+</span>
 
 <style type="text/scss">
   @import './node_modules/bulma/sass/utilities/_all';
-  // @import "./node_modules/bulma/sass/elements/tag.sass";
-
-  .tags .tag:not(:last-child) {
-    margin-right: .5rem;
-  }
-
-  .tags .tag {
-      margin-bottom: .5rem;
-  }
 </style>
--- a/workspaces/lluis/Tag.svelte
+++ b/workspaces/lluis/Tag.svelte
@@ -1,0 +1,37 @@
+<script lang="typescript">
+  export let isContainer = false
+  export let color = null
+  export let theme = null
+  export let size = null
+  export let isDelete = false
+  export let isRounded = false
+  export let style = ""
+</script>
+
+{#if isContainer}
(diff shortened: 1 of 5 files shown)


workspaces/web/src/routes/development.svelte
```diff
@@ -3,6 +3,8 @@
   import Mascot from "../components/Mascot"
   import TwitterButton from "../components/TwitterButton"
   import GitHubButton from "../components/GitHubButton"
+  import Tag from "lluis/Tag"
+  import Tag from "lluis/Tag"
```
kantord, line 7: This is a duplicate declaration, which is also making the CI fail:

https://kantord.semaphoreci.com/jobs/8d0c9eb7-efd9-437c-9b47-8286f390701e

workspaces/lluis/Tag.svelte
```diff
@@ -0,0 +1,37 @@
...
+  export let isContainer = false
+  export let color = null
+  export let theme = null
+  export let size = null
+  export let isDelete = false
+  export let isRounded = false
+  export let style = ""
+</script>
+
+{#if isContainer}
```
kantord, line 11: I wouldn't go with isContainer.

It seems like the container version has basically noting in common with the tag version.

So I'd rather have separate `<Tag />` and `<Tags />` components, which would simplify both the code and the usage

kenanchristian: This is now implemented in the new commit

workspaces/web/src/components/ChipsChallenge/index.svelte
```diff
@@ -116,7 +117,7 @@
         use:sortable="{{ items: answer, options: { group: 'chips', forceFallback: true } }}">
         {#each get(answer) as chip}
           <span class="chip">
-            <spain class="tag is-medium">{chip}</spain>
```
kantord, line 119: oh wow, what a typo that was :smile_cat:

workspaces/lluis/Tag.svelte
```diff
@@ -0,0 +1,21 @@
...
+  export let isDelete = false
+  export let isRounded = false
+  export let style = ""
+</script>
+
+<span {style} class="tag {`tag ${size ? `is-${size}`: ''} ${theme ? `is-${theme}`: ''} ${color ? `is-${color}`: ''}`}"
+class:is-rounded={isRounded}
+class:is-delete={isDelete}
+>
+  {#if !isDelete}
```
kantord, line 14: hmm, I'm not sure why this `isDelete` is necessary here :thinking: I think we're not really using this, no?

You can freely delete anything that we're not really using. Lets not maintain unused code :-)

kantord:
https://percy.io/LibreLingo/LibreLingo/builds/7524506?utm_campaign=LibreLingo&utm_content=LibreLingo&utm_source=github_status_public <----- Not sure if you've noticed it, but a couple things are visually broken. It looks like there's not any huge mistakes though.

Almost LGTM, just please fix the things that break visually, and remove any code that we're not actually using. :pray:

kantord:
Hi @kenanchristian , how is it going here?

kantord:
Feel free to re-open if you intend to finish it. Feel free to reach out if you need help

PR #872, opened 2020-10-30
feat: implement image component

### LINKED ISSUE #854: Turn sass/elements/image.sass into a component
This file is currently imported in [mystyles.scss](https://github.com/kantord/LibreLingo/blob/master/workspaces/web/src/mystyles.scss). It corresponds to a [bulma feature](https://bulma.io/documentation/).

This is all and well, but long term we want to avoid using bulma, and instead implement the styles within each component. The advantage of that is that we don't have to put unused CSS on pages, increasing our performance.

For now, this task consists of

- Creating a new [lluis component using Svelte](https://github.com/kantord/LibreLingo/tree/master/workspaces/lluis) that implements the same feature this sass file does
- Replace all usages of the classes found in this sass file with the component you just implemented.

**You don't need to re-implement the styles**. For now, it's perfectly fine if you just import the same .sass file, but only in the newly creating component. This already helps to achieve this optimization.

If you run into a tricky situation, don't be afraid to ask for help!

This is a very simple example of a component you might have to create:

https://github.com/kantord/LibreLingo/blob/9a84e20e6a14b0de6d31037039fcc5990dc4ed2f/workspaces/lluis/Box.svelte#L1-L3

As you can see, the main idea is implementing a simple component that adds a class.

If the usages involving this component are a little bit more tricky, you might have to introduce some props, as witnessed in this component: https://github.com/kantord/LibreLingo/blob/master/workspaces/lluis/Column.svelte

**No need to implement any component that isn't actually used.**

kenanchristian:
Implement Image Component (Fixes #854)

### DIFF
--- a/workspaces/lluis/Figure.svelte
+++ b/workspaces/lluis/Figure.svelte
@@ -1,18 +1,15 @@
 <script lang="typescript">
-  export let src
-  export let alt
   export let size
   export let className
   export let isRounded = false;
 </script>
 
-<img {src} {alt} class="image { size ? `is-${size}` : '' } {className}" class:is-rounded={isRounded}>
+<figure class="image { size ? `is-${size}` : '' } {className}" class:is-rounded={isRounded}>
+  <slot/>
+</figure>
+
 
 <style type="text/scss">
   @import './node_modules/bulma/sass/utilities/_all';
   @import './node_modules/bulma/sass/elements/image.sass';
-
-  .screenshot {
-    box-shadow: 0 0 2em #ffffff42;
-  }
 </style>
--- a/workspaces/lluis/Image.svelte
+++ b/workspaces/lluis/Image.svelte
@@ -1,0 +1,21 @@
+<script lang="typescript">
+  export let src
+  export let alt
+  export let size
+  export let className
+  export let isRounded = false;
+</script>
+
+<figure class="image { size ? `is-${size}` : '' } {className}" class:is-rounded={isRounded}>
+  <img {src} {alt}>
+</figure>
+
+
+<style type="text/scss">
+  @import './node_modules/bulma/sass/utilities/_all';
+  @import './node_modules/bulma/sass/elements/image.sass';
+
+  .screenshot {
+    box-shadow: 0 0 2em #ffffff42;
+  }
+</style>
--- a/workspaces/web/src/components/OptionCard.svelte
+++ b/workspaces/web/src/components/OptionCard.svelte
@@ -1,5 +1,5 @@
 <script lang="typescript">
-  import Image from "lluis/Image"
+  import Figure from "lluis/Figure"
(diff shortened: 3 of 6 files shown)


kantord:
There is a conflict in workspaces/web/src/routes/development.svelte

workspaces/lluis/Image.svelte
```diff
@@ -0,0 +1,21 @@
+<script lang="typescript">
+  export let src
+  export let alt
```
kantord, line 3: This could be an issue. Afaik there's no error when it's not provided. It should be a huge red flag thought, becase img tags need to have an alt value. You don't need to solve this issue in this MR, but once this MR is merged can you please create a new issue that asks for creating an eslint rule, which would flag usages of `lluis/Image` that don't have an alt tag provided?

workspaces/lluis/Image.svelte
```diff
@@ -0,0 +1,21 @@
...
+  export let size
+  export let className
+  export let isRounded = false;
+</script>
+
+<figure class="image { size ? `is-${size}` : '' } {className}" class:is-rounded={isRounded}>
```
kantord, line 9: I would be wary of always putting a figure tag around images. I think this isn't considered valid practice in HTML, as the figure tag has its own meaning.

Before this, `<figure />` was only added when that image was actually illustrating something :thinking: 

I would rather create a separate `<Figure />` component

workspaces/lluis/Image.svelte
kenanchristian: Okay got it 👌

workspaces/lluis/Image.svelte
kenanchristian: Implemented this in the new commit :)

workspaces/lluis/Figure.svelte
```diff
@@ -0,0 +1,15 @@
+<script lang="typescript">
+  export let size
+  export let className
+  export let isRounded = false;
```
kantord, line 4: Not sure if isRounded is really used anywhere. Please remove any code that's not really used. We don't want to maintain unused code :-)

workspaces/lluis/Image.svelte
kantord: Thanks!

kantord:
Please resolve the conflict in development.svelte

kantord:
Screenshot_2020-11-01 LibreLingo
 
the images don't seem to be displayed at all

kantord:
Screenshot_2020-11-01 LibreLingo(1)

these seem to be off too

kantord:
Hi @kenanchristian , how is it going? Do you need some help here?

kantord:
Feel free to re-open if you intend to finish it. Feel free to reach out if you need help

PR #1253, opened 2021-04-27
fix: correct path for the binary building pipeline

### LINKED ISSUE #1175: Create a very easily installable binary version
**Is your feature request related to a problem? Please describe.**
Sometimes a school or other organization might want to set up their own local server. They need an easy way of installing a server that doesn't involve having to read the documentation.

**Describe the solution you'd like**
- The solution should be cross-platform
- You should not need to have nodejs/npm/etc installed in order to install the binary version

Some tech information:

Here's how you export the whole page into a static site: https://github.com/kantord/LibreLingo/blob/main/.semaphore/pipeline_2.yml#L20

This is already done there in the CI, so basically this package could be built in the same CI step, utilizing the already existing static site.

A very simple, no-setup static site server can be used to serve these files on a local network. Here's one example: https://www.npmjs.com/package/serve

So this issue is mostly about figuring out how to package those things in a way that it becomes very easy to install, or even just a single executable on windows

dhelmr:
## Description

This should fix the failing pipeline for the librelingo-standalone-server builds. I forgot to modifiy the path in the pipeline configuration before, sorry.

fixes #1175

### DIFF
--- a/.semaphore/pipeline_2.yml
+++ b/.semaphore/pipeline_2.yml
@@ -32,6 +32,6 @@ blocks:
             - sem-version go 1.16
             - checkout
             - cache restore website-build-$SEMAPHORE_GIT_PR_SHA
-            - apps/librelingo-server/build.sh __sapper__/export/ binaries/
+            - apps/librelingo-standalone-server/build.sh __sapper__/export/ binaries/
             - artifact push project --force binaries      
 


kantord:
thanks, let see what happens if I merge it!

kantord:
hmm, seems like artifacts are empty :thinking: https://kantord.semaphoreci.com/artifacts/jobs/a7c1e445-5124-4833-bb97-98d759c37e72?project=dfeebc72-7f5b-4b25-ab20-831a2a2f6e07 cc. @dhelmr

dhelmr:
Hm, unfortunately, I cannot access this site. The links suggests that those are the job artifacts, but the pipeline should push the artifacts on the project level.

In my private semaphore project, I can access them at the project page (see screenshot):
grafik

kantord:
Very nice @dhelmr ! Indeed I can find it there!

PR #1294, opened 2021-05-13
chore(docs): Link to chat via matrix.to, instead of riot.im

binyamin:
Little aesthetics for the readme :grin:
Basically, the matrix.to service allows the user to choose which matrix client to open, whereas riot.im just opens the web app for Element (formerly riot).

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@
 </p>
 
 <p align="center">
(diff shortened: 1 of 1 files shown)


kantord:
Nice!

Thank you @binyamin !

@allcontributors please add @binyamin for docs

binyamin:
Thanks!

PR #1356, opened 2021-06-06
small typo fix

Lamdarer:
I think this shouldn't be "--ammend" just "--amend" could confuse a bit when following the Setup.


kantord:
thank you, @Lamdarer !

Issue #1450, opened 2021-07-15
Multi-word mini-dictionary terms


zmbc:
Right now, each *word* in a `Translation` must have a definition in a `Mini-dictionary`. In other words, the `Translation`s are split on the space character to break them into chunks.

But this is not always the best way to split them, pedagogically. Certain phrases are more relevant to the lesson at the phrase level -- e.g. "post office" or "land animal" or "blue whale" -- especially when that concept is represented in the target language by a single word, or a completely different compound phrase. In that case, being able to look up the words "blue" and "whale" separately is no help at all.

Therefore, a useful feature would be able to manually override the behavior to join certain multi-word terms into single mini-dictionary entries. @kantord [suggested](https://github.com/kantord/LibreLingo/issues/1445#issuecomment-880815451) curly braces around the phrase, e.g. `{post office}`. These would be underlined together, and mousing over them would bring up a single tooltip with the corresponding translation of the phrase.

I am happy to tackle implementing this feature.

kantord:
I think this can be implemented relatively easily by changing this line:

https://github.com/kantord/LibreLingo/blob/0448295e71fd2e0b8b6190f067c90a0e80586f3b/apps/librelingo_json_export/librelingo_json_export/dictionary.py#L63

Although it's possible that other lines will have to be changed as well, in order to make sure that the curly braces don't show up in the final files.

Also, it probably makes sense to also make it so that `{post office}` will become a single chip in chip challenges? In that case, you'll also have to change the implementation here: https://github.com/kantord/LibreLingo/blob/main/apps/librelingo_json_export/librelingo_json_export/challenge_types.py#L71

PR #1452, opened 2021-07-17
feat: Add support for single words/terms with spaces in them

zmbc:
Implements the feature described in #1450.

Note that this does not make any changes to the chips (they will still split on spaces). Personally, I don't see the value in joined chips, but if it's something you're set on, it can always be implemented separately.

### DIFF
--- a/apps/librelingo_json_export/librelingo_json_export/challenge_types.py
+++ b/apps/librelingo_json_export/librelingo_json_export/challenge_types.py
@@ -1,10 +1,7 @@
 import editdistance  # type: ignore
 
-from librelingo_utils import get_dumb_opaque_id, audio_id, clean_word, iterate_phrases
+from librelingo_utils import get_dumb_opaque_id, audio_id, remove_control_characters_for_display, clean_word, iterate_phrases
 from .dictionary import _define_words_in_sentence
-
-def _remove_special_characters_for_display(string):
-    return string.replace('}', '').replace('{', '')
 
 def get_listening_challenge(source, course):
     if not course.settings.audio_files_enabled:
@@ -13,9 +10,9 @@ def get_listening_challenge(source, course):
     return [
         {
             "type": "listeningExercise",
-            "answer": _remove_special_characters_for_display(source.in_target_language[0]),
-            "meaning": _remove_special_characters_for_display(source.in_source_language[0]),
-            "audio": audio_id(course.target_language, _remove_special_characters_for_display(source.in_target_language[0])),
+            "answer": remove_control_characters_for_display(source.in_target_language[0]),
+            "meaning": remove_control_characters_for_display(source.in_source_language[0]),
+            "audio": audio_id(course.target_language, remove_control_characters_for_display(source.in_target_language[0])),
             "id": get_dumb_opaque_id("Word", source, "listeningExercise"),
             "priority": 1,
             "group": get_dumb_opaque_id("Group", source),
@@ -30,7 +27,7 @@ def get_short_input_challenge(source, course):
             "pictures": [pic + ".jpg" for pic in source.pictures]
             if source.pictures
             else None,
-            "formInTargetLanguage": [_remove_special_characters_for_display(t) for t in source.in_target_language],
+            "formInTargetLanguage": [remove_control_characters_for_display(t) for t in source.in_target_language],
             "phrase": _define_words_in_sentence(
                 course,
                 source.in_source_language[0],
@@ -50,8 +47,8 @@ def get_cards_challenge(word, _):
             "pictures": [pic + ".jpg" for pic in word.pictures]
             if word.pictures
             else None,
-            "formInTargetLanguage": _remove_special_characters_for_display(word.in_target_language[0]),
-            "meaningInSourceLanguage": _remove_special_characters_for_display(word.in_source_language[0]),
+            "formInTargetLanguage": remove_control_characters_for_display(word.in_target_language[0]),
+            "meaningInSourceLanguage": remove_control_characters_for_display(word.in_source_language[0]),
             "id": get_dumb_opaque_id("Word", word, "cards"),
             "priority": 0,
             "group": get_dumb_opaque_id("Group", word),
@@ -63,8 +60,8 @@ def get_options_challenge(phrase, _):
     return [
         {
             "type": "options",
-            "formInTargetLanguage": _remove_special_characters_for_display(phrase.in_target_language[0]),
-            "meaningInSourceLanguage": _remove_special_characters_for_display(phrase.in_source_language[0]),
+            "formInTargetLanguage": remove_control_characters_for_display(phrase.in_target_language[0]),
+            "meaningInSourceLanguage": remove_control_characters_for_display(phrase.in_source_language[0]),
             "id": get_dumb_opaque_id("Options", phrase, "options"),
             "priority": 0,
             "group": get_dumb_opaque_id("Group", phrase),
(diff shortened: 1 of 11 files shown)


apps/librelingo_json_export/pyproject.toml
```diff
@@ -18,7 +18,7 @@ editdistance = "^0.5.3"
 pytest = "^6.2.2"
 pytest-watch = "^4.2.0"
 pyfakefs = "^4.3.3"
-librelingo-fakes = "^3.0.0"
+librelingo-fakes = {path = "../librelingo_fakes", develop = false}
```
zmbc, line 21: For me anyway, this seems to make testing a lot easier. If this is bad in some way, I can revert.

kantord: Good idea!

apps/librelingo_json_export/tests/test_export.py
```diff
@@ -491,6 +512,22 @@ def test_matches_definitions_in_a_case_insensitive_way(self):
             "definition": "by a lot",
         }
 
+    def test_matches_definitions_with_spaces(self):
+        my_course = Course(
+            **{
+                **(fakes.course1._asdict()),
```
kantord, line 518: maybe `fakes.customize` would make this easier/simpler? What do you think about it?

apps/librelingo_json_export/tests/test_export.py
```diff
@@ -423,6 +439,11 @@ def test_defines_every_word(self, _define_word):
             _define_word.return_value,
         ]
 
+        assert _define_words_in_sentence(fakes.course1, "{foo bar} {baz quux}", True) == [
```
kantord, line 442: please introduce a new test case for this one

apps/librelingo_json_export/tests/test_export.py
```diff
@@ -401,20 +401,36 @@ def test_calls_define_word_the_correct_number_of_times(self, _define_word):
...
+        _define_word.assert_called_with(fakes.course1, fake_word, is_in_target_language)
+
     @patch("librelingo_json_export.dictionary._define_word")
     def test_returns_correct_value(self, _define_word):
         _define_word.return_value = fakes.fake_value()
         assert _define_words_in_sentence(fakes.course1, "foo", True) == [
             _define_word.return_value
         ]
 
+        assert _define_words_in_sentence(fakes.course1, "{foo bar}", True) == [
```
kantord, line 430: same as https://github.com/kantord/LibreLingo/pull/1452/files#r671887184

apps/librelingo_json_export/tests/test_export.py
```diff
@@ -401,20 +401,36 @@ def test_calls_define_word_the_correct_number_of_times(self, _define_word):
...
+
     @patch("librelingo_json_export.dictionary._define_word")
     def test_calls_define_word_with_the_correct_data(self, _define_word):
         is_in_target_language = fakes.fake_value()
         fake_word = str(fakes.fake_value())
         _define_words_in_sentence(fakes.course1, fake_word, is_in_target_language)
         _define_word.assert_called_with(fakes.course1, fake_word, is_in_target_language)
 
+    @patch("librelingo_json_export.dictionary._define_word")
+    def test_calls_define_word_with_the_correct_data_with_spaces(self, _define_word):
```
kantord, line 417:
```suggestion
    def test_calls_define_word_with_the_correct_word_that_has_space(self, _define_word):
```

kantord: I am thinking you can probably merge this test case with this other one: https://github.com/kantord/LibreLingo/pull/1452/files#diff-ab9c50ea4369e0645fb5e151128d58dde9e38e4879661c5890f0be840cb5e650R404

You could basically assert the full list of calls, which would both assert that the call arguments are correct, as well as that they were called the correct number of times.

Something like this maybe? https://stackoverflow.com/a/24063759

apps/librelingo_json_export/librelingo_json_export/dictionary.py
```diff
@@ -60,4 +61,5 @@ def _define_words_in_sentence(course, sentence, reverse):
     """
     Converts a sentence into a list of definition objects.
     """
-    return [_define_word(course, word, reverse) for word in sentence.split()]
+    words = [p for p in re.split("( |\\{.*?\\}|'.*?')", sentence) if p.strip(' {}')]
```
kantord, line 64: I would really extract this to a new function, something like `parse_phrase` that could probably be part of librelingo-utils, so that it can be shared between multiple modules, not just json-export

apps/librelingo_json_export/librelingo_json_export/challenge_types.py
```diff
@@ -3,6 +3,8 @@
 from librelingo_utils import get_dumb_opaque_id, audio_id, clean_word, iterate_phrases
 from .dictionary import _define_words_in_sentence
 
+def _remove_special_characters_for_display(string):
+    return string.replace('}', '').replace('{', '')
```
kantord, line 7: same as here: https://github.com/kantord/LibreLingo/pull/1452/files#r671888194

and while at it, I would probably rename it to something more generic, like `_format_text_for_display`, or whatever that's better that you can think of. The reason why I say this is because it's possible that soon new special characters will have to be removed, so it will require less work to introduce those if the function name doesn't have to be changed

kantord (changes requested):
Hi, sorry for the slow review. I was a bit busy.

Good job on the merge request 💯 🚀

I have some small comments, please check them out and push fixes before we can merge it 😉

kantord:
Also look at the pipeline failure: https://kantord.semaphoreci.com/jobs/1ebb03c0-d22c-4f77-9cd7-965200363136

apps/librelingo_json_export/tests/test_export.py
zmbc: I copied this from [the code directly above](https://github.com/kantord/LibreLingo/pull/1452/files#diff-ab9c50ea4369e0645fb5e151128d58dde9e38e4879661c5890f0be840cb5e650L481). If you'd like me to do it both places, I can switch to `fakes.customize`.

apps/librelingo_json_export/tests/test_export.py
zmbc: Here also, I copied this structure from the surrounding tests: [here](https://github.com/kantord/LibreLingo/pull/1452/files#diff-ab9c50ea4369e0645fb5e151128d58dde9e38e4879661c5890f0be840cb5e650L400) and [here](https://github.com/kantord/LibreLingo/pull/1452/files#diff-ab9c50ea4369e0645fb5e151128d58dde9e38e4879661c5890f0be840cb5e650L405).

If you'd like me to merge those two as well, I can take a crack at it.

apps/librelingo_json_export/tests/test_export.py
kantord: > I copied this from the code directly above. If you'd like me to do it both places, I can switch to fakes.customize.

sounds good, but I'm also happy to hear your opinion about the usefulness of that function. Such opinions are very valuable for the project, as I'm not getting many code reviews myself 😸

apps/librelingo_json_export/tests/test_export.py
kantord: > If you'd like me to merge those two as well, I can take a crack at it.

sounds good to me. If you do that, I'll be very grateful.

zmbc:
@kantord I think I've addressed everything.

kantord:
@zmbc please look at the semaphore pipeline failure: https://kantord.semaphoreci.com/workflows/65f6c70c-60b3-4c29-a663-3c3be01b2ba0

zmbc:
@kantord Apologies for the delay.

The issue was that I was using a relative-path-based install for librelingo-utils (which is a regular dependency, not a dev dependency). I have reverted that change.

Correct me if I'm wrong, but now the librelingo_json_exporter package needs to use a PyPI version of librelingo-utils. So because this change to librelingo_json_exporter depends on my changes to librelingo-utils, I will need you to push version 2.4.2 from my branch to PyPI in order to use it in librelingo_json_exporter? Or is there another way?

kantord:
@zmbc that's correct. I don't know of any other way either.

kantord:
Thanks @zmbc!!!! 🎉 💯

One last thing, can you please create another MR to update the [course-creator documentation ](https://github.com/kantord/LibreLingo/blob/main/docs/courses/skill.md#data-breakdown)to document this new feature? 🙏

kantord:
@allcontributors please add @zmbc  for code

PR #1482, opened 2021-07-27
fix: do not display multiple of the same option in options challenges

zmbc:
Prevents the user from seeing multiple copies of the same option in an options challenge.

### DIFF
--- a/apps/web/src/courses/test/challenges/cards-test.json
+++ b/apps/web/src/courses/test/challenges/cards-test.json
@@ -51,11 +51,11 @@
         "pasta1.jpg",
         "pasta1.jpg"
       ],
-      "formInTargetLanguage": "pasta",
+      "formInTargetLanguage": "pasta1",
       "meaningInSourceLanguage": "pasta__1",
-      "id": "573e78d0b0ca",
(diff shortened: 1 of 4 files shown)


kantord (approved):

kantord (changes requested):
Hi @zmbc, thanks for the contribution.

Looks like the eslint is failing: https://kantord.semaphoreci.com/jobs/586e351b-ca09-4c64-b629-40aaaf7b4b4d

Also, it looks like the Cypress tests might need some updating: https://kantord.semaphoreci.com/jobs/48cf3499-c5b0-419b-ac0d-3e0420668743

(Or maybe your change introduces some unintended side-effect?)

zmbc:
@kantord Both issues should be resolved now.

By the way, this PR doesn't totally fix the problem, because it could still happen that an option is given that matches an *alternative* translation of the phrase/word. However, this info isn't even exported in the JSON yet so I have chosen not to tackle that here. Perhaps it makes sense to create an open issue to track this?

kantord:
Thank you @zmbc ! 🏅

Issue #1445, opened 2021-07-13
toki pona


zmbc:
Hi!

I am part of a community of toki pona speakers who are looking for a new way to teach toki pona.

[toki pona](https://en.wikipedia.org/wiki/Toki_Pona) is a minimalist conlang (constructed language) with a small but vibrant community.

If we were to create a toki pona course, would you consider accepting it into LibreLingo?

If not, would it be alright for us to fork the project and create a separate toki pona platform (removing the name LibreLingo)?

Thanks!

kantord:
of course, I would accept it into LibreLingo!

zmbc:
@kantord Great! I will update you on progress

kantord:
Amazing! Let me know if you need help. I'm also happy testing :smile_cat:

zmbc:
@kantord I have a question about the mini-dictionary. As far as I can tell, it is used when the user hovers over an underlined word, which can occur in short-answer word questions or when translating a phrase. It's required for every word that appears in translations.

There are some situations where this feels awkward. Two come to mind:
- If a `Translation` is more than one word, an entry is required for each component word. If a word means "post office," you need entries for "post" and "office" in the mini-dictionary, which probably don't have anything to do with the lesson.
- If a word has no clear equivalent. For example, if the source language has articles (like English) and the target language has none (like Japanese, Korean, or toki pona), there is not going to be a target language definition/equivalent for them.

What do you think about making mini-dictionary entries optional?

(Note: I am happy to do the implementation of any new features in LibreLingo required for toki pona. I just want to ask you this from a design perspective.)

kantord:
> If a Translation is more than one word, an entry is required for each component word. If a word means "post office," you need entries for "post" and "office" in the mini-dictionary, which probably don't have anything to do with the lesson.

This could be solved by adding an optional way to separate words. For example instead of of writing:

```
Marie is going to the post office
```

you'd write

```
Marie is going to the {post office}
```

and then "post office" would be handled as a single word, but the `{}` characters would be dropped in the final result

Can you please create a separate issue for it? :pray:

kantord:
> If a word has no clear equivalent. For example, if the source language has articles (like English) and the target language has none (like Japanese, Korean, or toki pona), there is not going to be a target language definition/equivalent for them.

[...]

PR #1494, opened 2021-08-01
fix: Accept solutions to chips challenges regardless of case

zmbc:
As mentioned [here](https://github.com/kantord/LibreLingo/issues/1445#issuecomment-890482465), requiring correct capitalization for chips challenge solutions is not intended. This change implements that, so that "como pan" is correct if there is an alternative solution "Como pan".


kantord:
Amazing! 💯

thank you @zmbc

PR #1495, opened 2021-08-02
docs: Document challenge types and when they are generated

zmbc:
Following up on [conversation](https://github.com/kantord/LibreLingo/issues/1445#issuecomment-890026683) in #1445, this PR adds documentation for each type of challenge, when it is generated, and what properties of the YAML configuration it uses.


docs/courses/README.md
```diff
@@ -43,11 +43,12 @@ You'll need to have an idea on basic terminology used in LibreLingo:
 - A **Course** is a set of Modules. There is typically one course for each language pair. e.g. Spanish for English speakers, Japanese for Italian speakers, etc.
 - A **Module** is a set of Skills.
 - A **Skill** is a small unit of course material. In the web app, the user will typically practice a Skill at a time.
+- A **Challenge** is a single question or task the user will complete within a skill.
```
kantord, line 46: hmm, I think Course/Module/Skill *vs* Challenge are not really comparable as they are not the same category.

Course/Module/Skill are categories within the YAML based course system. Challenges are a different concept that are specific to the web app.

The distinction is that the YAML format is meant to be a generic format for building courses, whereas challenges live in the JSON format which is a different layer that is used to implement the web app. This might still be relevant to course authors, but the fact that there's no such thing as "Challenge" in the YAML files is important.

(The other changes in this file make sense)

docs/courses/skill.md
```diff
@@ -108,17 +111,18 @@ Mini-dictionary:
...
 
-**`New words`** has a list of new words that the lesson asks about.
-- `Word`: The word in destination language.
+**`New words`** has a list of new words that the lesson teaches.
+- `Word`: The word in the target language, i.e. the language the user is learning.
 - `Synonyms`: A list of synonyms of the above word. (optional)
-- `Translation`: Translation of the word in target audience's language.
-- `Also accepted`: A list of synonyms of the `Translation` word above. (optional)
+- `Translation`: Translation of the word in the source language, i.e. the language the user already knows.
+- `Also accepted`: A list of alternative translations to the source language, which will *not* be taught, but will be accepted as correct translations of the `Word` field if supplied by the user. Currently has no effect, but may be used by future challenge types. (optional)
```
kantord, line 118: for words, does this really have no effect? I think in short input challenges, users are asked to type out the translation of a word, so I think it's already used 🤔 _Phrases_ don't have this, but if I remember correctly, _Words_ do

```suggestion
- `Also accepted`: A list of alternative translations to the source language, which will *not* be taught, but will be accepted as correct translations of the `Word` field if supplied by the user. (optional)
```

docs/courses/skill.md
```diff
@@ -131,6 +135,20 @@ Note that all `New words` are automatically added to the `Mini-dictionary`, in b
 - `<destination language>`: A list of terms in destination language as key and meaning in target audience's language.
 - `<target audience's language>`: A list of terms in target audience's language as key and meaning in destination language.
 
+## How skills generate challenges
```
kantord, line 138:
```suggestion
## How skills are displayed to the user
```

docs/courses/skill.md
```diff
@@ -131,6 +135,20 @@ Note that all `New words` are automatically added to the `Mini-dictionary`, in b
...
+
+As the course author, you only have to specify the words and phrases you want to teach, and LibreLingo will generate the challenges (individual questions). However, it is useful to understand what challenges will be generated.
```
kantord, line 140:
```suggestion
As the course author, you have to specify the words and phrases you want to teach. LibreLingo will know how to present skills as a series of individual questions or challenges. It is useful to understand what challenges will be generated.
```

docs/courses/challenge.md
```diff
@@ -0,0 +1,87 @@
+# LibreLingo Documentation on Challenge
+
+A **Challenge** is a single task or question the user responds to as they study using LibreLingo. Challenges are not configured directly, but [automatically generated from Skill configuration](skill.md#how-skills-generate-challenges).
```
kantord, line 3:
```suggestion
As they study using LibreLingo, users respond to a series of tasks or questions called **Challenge**s.

Challenges are not configured directly, but [automatically generated from Skill configuration](skill.md#how-skills-generate-challenges).
```

(Also updated the link: skill.md#how-skills-generate-challenges accordingly)

docs/courses/challenge.md
```diff
@@ -0,0 +1,87 @@
...
+
+To get a better understanding, you can read the [course basics](README.md#basics).
```
kantord, line 5: to get a better understanding of what?

docs/courses/challenge.md
```diff
@@ -0,0 +1,87 @@
...
+**Table of Contents:**
+- [Cards challenge](#cards-challenge)
+- [Short input challenge](#short-input-challenge)
+- [Options challenge](#options-challenge)
+- [Chips challenge](#chips-challenge)
+- [Listening challenge](#listening-challenge)
+
+## Cards challenge
+
+<img src="/docs/screenshots/cards-challenge.png" width="400px"  alt="Text asks the user: which of these is \"river\"?. Below this heading are three cards. Each card contains a Spanish word and corresponding image. In the center the correct choice (río) is selected."></img>
```
kantord, line 23: Good job for writing a good alt text and including images! 💯

I wonder though, is it possible to use alt texts with "proper" markdown syntax? 🤔 https://github.github.com/gfm/#images

kantord:
Otherwise LGTM, very good job! 💯

docs/courses/skill.md
zmbc: Unless I'm mistaken, short input challenges only ever prompt the user with the source language, and ask the user to type an answer in the target language. Therefore, either the `Word` or any of its `Synonyms` would be correct, but the `Also accepted` field is never used, as the prompt is always the `Translation`.

docs/courses/README.md
zmbc: Yeah, you're right. I wanted to define the word Challenge somewhere, but this isn't a good place. I'll try a different way, and we can see if that's better.

docs/courses/challenge.md
zmbc: :laughing: I don't know, I think I copied this from another page and forgot to delete it.

docs/courses/challenge.md
zmbc: I'll play around with it -- I don't think `width="400px"` is supported by GFM but that might not even be necessary.

zmbc:
@kantord Think I've resolved everything.

kantord (approved):

kantord:
Thank you @zmbc !

Issue #1473, opened 2021-07-23
Easier-to-use audio generation


zmbc:
**Is your feature request related to a problem? Please describe.**
If I want to add a new course with TTS listening exercises, it is not easy to do so.

**Describe the solution you'd like**
I am not sure about all the details, but it would at least include:
- Configuration of TTS in the course YAML file.
- A script that recognizes missing audios, generates them using the course's TTS configuration, and places them in the right directory with the right names.

**Describe alternatives you've considered**
I've seen [this](https://github.com/kantord/LibreLingo/blob/main/docs/creating_a_python_language_learning_app_using_librelingo_courses.md) but it's more of an example for developers than anything, and in any case it doesn't show the whole process.

The fetchAudios.sh script is in bash, which is probably not ideal and even less ideal if it gets more complex. I'm guessing Python would be better. There is also no documentation I can find of how the audios_to_fetch.csv file is generated.

**Additional context**
I'm interested in implementing this, if it gets the go-ahead. If it's not too much to review, I think this and allowing other audio sources could be combined into one MR.

kantord:
@zmbc see also:

- https://github.com/kantord/LibreLingo/issues/1464
- https://github.com/kantord/LibreLingo/issues/714

those are possible alternatives to using TTS.

Of course I will also accept your suggested contribution 😉 How do you plan to implement it?

zmbc:
@kantord Thanks, that's helpful.

Here's my initial idea of how this would work.

*course.yaml*
```yaml
...
Settings:
  Audio:
    Enabled: True
    TTS:
      # Enabled: False can be used if audio should only be present on words/phrases with recordings
      Enabled: True # Default: True
      Provider: Polly # Default: Polly, can be used in future if other TTS providers are available
      Voice: Aditi # No default
      Engine: standard # Can be standard or neural, default: standard
```

*skill.yaml*
```yaml
New words:
  # This word will feed the `Word` field into the course's TTS engine
  - Word: soweli
    Translation: dog
    Images:
      - dog1
      - dog2
      - dog3
  # This word will pronounce a custom IPA string using the course's TTS engine (while still displaying the word as written, of course)
  - Word: soweli
    IPA: "'soweli"
    Translation: cat
    Images:
      - cat1
      - cat2
      - cat3
  # This word will use a recording from a third party
  - Word: soweli
    Audio source: Tatoeba # Default: TTS
[...]

PR #1528, opened 2021-08-13
feat: audio generation script: TTS only

zmbc:
## Description

With the changes made here, TTS can be configured on a per-course basis in YAML, and then a script can be used to generate audio using those settings. The script can run in either update (only generate new audios) or overwrite (generate everything from scratch) modes. Audios of removed phrases will be deleted automatically. An index JSON file is created for each course.

Note that this PR does not include the *results* of running the script for existing courses.

This is the first step of #1473. The further work that could be done:
- Add phrase-level setting to customize IPA
- Other audio sources & attribution
- Possible stylistic changes like creating `voice` subdirectories

I've used tons of relative path installs (otherwise it's not really possible to develop without write access to the PyPI packages), so I'll need help reverting those and replacing with actual published versions.

## Checklist

- [x] The CI is passing
- [x] I wrote new tests to test my code, or updated existing tests to test my changes


apps/librelingo_audios/librelingo_audios/generate_audios.py
```diff
@@ -0,0 +1,88 @@
...
+    if settings.overwrite:
+        # Pretend nothing already exists; recreate everything
+        phrase_indices_already_present = []
+    else:
+        with open(index_path, "r") as f:
+            phrase_indices_already_present = json.loads(f.read())["audios"]
+
+    result_index = []
+
+    # For each phrase in the course, if it doesn't exist, generate it.
```
kantord, line 24: What do you think about the idea of extracting this logic into a new function?

```python
result_index = _load_result_index()
```

It could basically combine this logic:

```python
        with open(index_path, "r") as f:
            phrase_indices_already_present = json.loads(f.read())["audios"]
```

with also the deduplication logic (at least I assume the logic below is for deduplication?)

apps/librelingo_types/librelingo_types/data_types.py
```diff
@@ -1,7 +1,63 @@
...
+        provider="Polly",
+        voice="Aditi",
+        engine="standard"
+    )
+    ```
+    """
+
+    pass
+
+default_text_to_speech_settings = TextToSpeechSettings()
```
kantord, line 30:
```suggestion
DEFAULT_TEXT_TO_SPEECH_SETTINGS = TextToSpeechSettings()
```

apps/librelingo_types/librelingo_types/data_types.py
```diff
@@ -1,7 +1,63 @@
...
+    AudioSettings(
+        enabled=True,
+        text_to_speech_settings=[TextToSpeechSettings()]
+    )
+    ```
+    """
+
+    pass
+
+default_audio_settings = AudioSettings()
```
kantord, line 57:
```suggestion
DEFAULT_AUDIO_SETTINGS = AudioSettings()
```

apps/librelingo_types/librelingo_types/data_types.py
```diff
@@ -1,7 +1,63 @@
...
+
 
-class Settings(namedtuple("Settings", ["audio_files_enabled"], defaults=[True])):
+class Settings(namedtuple("Settings", ["audio_settings"], defaults=[default_audio_settings])):
```
kantord, line 60: this is meant to be a breaking change, right?

apps/librelingo_types/librelingo_types/data_types.py
```diff
@@ -1,7 +1,63 @@
...
+default_text_to_speech_settings = TextToSpeechSettings()
+
+class AudioSettings(
+    namedtuple(
+        "AudioSettings",
+        [
+            "enabled",
+            "text_to_speech_settings",
+        ],
+        defaults=[True, [default_text_to_speech_settings]]
```
kantord, line 39:
```suggestion
        defaults=[False, []]
```

what do you think? Wouldn't it be better to have audio disabled by default? That way course authors would need to be more mindful in choosing how they set it up

For text to speech it will need manual setup anyway, no? 🤔

apps/librelingo_yaml_loader/librelingo_yaml_loader/yaml_loader.py
```diff
@@ -332,14 +334,49 @@ def _convert_license(raw_license):
...
+    Creates an AudioSettings() object based on the data structure in the YAML
+    file
+    """
+    if "Audio" not in raw_settings:
+        return AudioSettings()
+
+    raw_audio_settings = raw_settings["Audio"]
+
+    if raw_audio_settings["Enabled"]:
+        text_to_speech_settings = _convert_text_to_speech_settings(raw_audio_settings)
```
kantord, line 362:
```suggestion
        text_to_speech_configs = _convert_text_to_speech_settings(raw_audio_settings)
```

I feel like it's a better naming, no? Maybe throughout the codebase. If I understand correctly, there can be multiple configs simultaneously

apps/librelingo_yaml_loader/tests/test_yaml_loader.py
```diff
@@ -176,20 +176,99 @@ def test_returns_correct_special_characters(self):
...
+    def test_returns_correct_settings_audio_settings_enabled(self):
+        assert self.result.settings.audio_settings.enabled == True
 
-    def test_returns_correct_settings_audio_files_disabled(self):
+    def test_returns_correct_settings_audio_settings_tts(self):
+        tts_settings = self.result.settings.audio_settings.text_to_speech_settings
+        assert len(tts_settings) == 1
+        assert tts_settings[0].provider == "Polly"
+        assert tts_settings[0].voice == "Lupe"
+        assert tts_settings[0].engine == "standard"
```
kantord, line 187: How about merging it into a single assertion?

```suggestion
        assert tts_settings = [
          TextToSpeechSettings(
            provider="Polly",
            voice="Lupe",
            engine="standard",
          )
        ]
```

apps/librelingo_yaml_loader/tests/test_yaml_loader.py
```diff
@@ -176,20 +176,99 @@ def test_returns_correct_special_characters(self):
...
+
+    def test_returns_correct_settings_audio_disabled(self):
+        new_settings = """
+    Settings:
+        Audio:
+            Enabled: False
+        """
+
+        # Append settings to the file
```
kantord, line 196: this part seems to show up multiple times, maybe it could be extended like this?

```python
append_settings_to_file("""
    Settings:
        Audio:
            Enabled: False
        """)

```

apps/librelingo_yaml_loader/tests/test_yaml_loader.py
```diff
@@ -176,20 +176,99 @@ def test_returns_correct_special_characters(self):
...
+        with open(Path(self.fake_path) / "course.yaml", "a") as f:
+            f.write(new_settings)
+        self.result = load_course(self.fake_path)
+        assert self.result.settings.audio_settings.enabled == True
+
+        tts_settings = self.result.settings.audio_settings.text_to_speech_settings
+        assert len(tts_settings) == 0
```
kantord, line 242:
```suggestion
        assert tts_settings == []
```

apps/librelingo_audios/librelingo_audios/generate_audios.py
```diff
@@ -0,0 +1,88 @@
...
+def _generate_audio_for_phrase(phrase, output_path, course, settings):
+    tts_settings = random.choice(course.settings.audio_settings.text_to_speech_settings)
+    file_name = audio_id(course.target_language, phrase.in_target_language[0])
+    destination_path = Path(Path(output_path) / "{}.mp3".format(file_name))
+
+    if destination_path.is_file() and not settings.overwrite:
+        raise RuntimeError('File {} already exists!'.format(destination_path))
+
+    if settings.dry_run:
+        print('Would generate {} using {} {}'.format(destination_path, tts_settings.voice, tts_settings.engine))
```
kantord, line 57: I wonder, should dry run be handled globally or per provider?

Basically I now think it should be per-provider, because I think the dry-run should be as real as possible, so for example any validation should still be executed (but network calls or file operations should not be executed)

kantord:
> Note that this PR does not include the results of running the script for existing courses.

why did you decide to not include that? I think it would be worth including that for Spanish for example

apps/librelingo_audios/librelingo_audios/cli.py
```diff
@@ -0,0 +1,47 @@
...
+    if not settings.dry_run:
+        Path(output_path).mkdir(parents=True, exist_ok=True)
+
+
+@click.command()
+@click.argument("input_path")
+@click.argument("output_path")
+@click.argument("course_name")
+@click.option("--dry-run/--no-dry-run", default=DEFAULT_SETTINGS.dry_run)
+@click.option("--overwrite/--no-overwrite", default=DEFAULT_SETTINGS.overwrite)
```
kantord, line 32: I would put some `help=""` here explaining what this does, as the idea of downloading everything all over from scratch is a bit strong. Some people might believe that this would just _update_ things that need to be updated

apps/librelingo_audios/librelingo_audios/generate_audios.py
```diff
@@ -0,0 +1,88 @@
+from librelingo_utils import audio_id, iterate_phrases
+from pathlib import Path
+import subprocess
+import json
+import random
+
+def generate_audios_for_course(output_path, course_name, course, settings):
```
kantord, line 7:
```suggestion
def update_audios_for_course(output_path, course_name, course, settings):
```

since I think it doesn't necessarily always only generate, but potentially also:

- updates
- deletes
- downloads

no? 🤔

apps/librelingo_audios/librelingo_audios/generate_audios.py
```diff
@@ -0,0 +1,88 @@
...
+    if not course.settings.audio_settings.enabled:
+        return
+
+    index_path = Path(Path(output_path) / "{}.json".format(course_name))
+
+    phrases_in_course = [(_index_dict_for_phrase(p, course), p) for p in iterate_phrases(course)]
+
+    if settings.overwrite:
+        # Pretend nothing already exists; recreate everything
```
kantord, line 16: so this is effectively equivalent to deleting the meta-data file? By the way, how will this make sure that also the already existing audio files are deleted or overwritten?

apps/librelingo_audios/librelingo_audios/generate_audios.py
```diff
@@ -0,0 +1,88 @@
...
+        # multiple sources
+        existing_matching_index = next((i for i in phrase_indices_already_present if i["id"] == course_phrase_index["id"]), None)
+
+        if existing_matching_index is None:
+            _generate_audio_for_phrase(course_phrase, output_path, course, settings)
+            result_index.append(course_phrase_index)
+        else:
+            result_index.append(existing_matching_index)
+
+    # For phrases that already exist as audio, check if they are still needed by
```
kantord, line 36: shouldn't this happen before we are generating the audio?

Maybe we could have a test case where in the test setup, we have:

- (A) an audio file that has never been in the meta data and is not needed
- (B) an audio file that's in the meta data now, but the meta data is outdated and after updating it, the file is now not necessary
- (C) an audio file that is still needed and is correctly supplied
- (D) a missing audio file
- (E) an audio file that's needed and exists, but it's missing from the meta file

We can mock the audio file generation so that it always returns with a fake audio file that's based on the original phrase (for example, the content is sth like `f"fake audio file for phrase '{phrase_text}'"`

the assertions should be:

- (A) is deleted
- (B) is also deleted
- (C) remains unchanged
- (D) is generated using the fake audio generation function
- (E) is generated using the fake audio generation function, because do to the lack of data in the metadata, we need to suspect that it's an incorrect file or lacking license

apps/librelingo_audios/librelingo_audios/generate_audios.py
```diff
@@ -0,0 +1,88 @@
...
+    # the course.
+    for index in phrase_indices_already_present:
+        matching_course_phrase = next(((i, p) for i, p in phrases_in_course if i["id"] == index["id"]), None)
+
+        if matching_course_phrase is None:
+            _delete_audio_for_phrase(course_phrase, output_path, course, settings)
+
+    if not settings.dry_run:
+        with open(index_path, "w", encoding="utf-8") as f:
+            json.dump({"audios": result_index}, f, ensure_ascii=False, indent=4)
```
kantord, line 46: Why not?

```suggestion
            json.dump(result_index, f, ensure_ascii=False, indent=4)
```

kantord: Also why ensure_ascii=False?

apps/librelingo_audios/librelingo_audios/generate_audios.py
```diff
@@ -0,0 +1,88 @@
...
+        return
+
+    if settings.dry_run:
+        print('Would delete {}'.format(target_path))
+    else:
+        print('Deleting {}'.format(target_path))
+        target_path.unlink()
+
+def _index_dict_for_phrase(phrase, course):
+    # TODO add more fields to indicate sources other than TTS
```
kantord, line 84: I would already add some files:

- license, since audios generated by the TTS still need to be licensed. If the TTS service guarantees you the right to license the resulting audio files as you like, then it is probably fine to make it default to the license of the course itself

kantord:
> > Note that this PR does not include the results of running the script for existing courses.
> 
> why did you decide to not include that? I think it would be worth including that for Spanish for example

I think this would also be useful because we'd essentially be able to manually integration test that everything just works

kantord:
@zmbc I left some comments/questions. Thank you so much for this PR, it looks amazing 🎉 Good job

apps/librelingo_types/librelingo_types/data_types.py
zmbc: It is breaking in that the format of the `course.yaml` file changes, yes. Also, because I agree with your idea in the other comment about audio being off by default, it will be breaking in changing the default.

apps/librelingo_audios/librelingo_audios/generate_audios.py
zmbc: Good idea. I'm actually refactoring the whole method to make it clearer. The logic below was not for deduplication, so clearly it could use some refactoring :)

apps/librelingo_audios/librelingo_audios/generate_audios.py
zmbc: Agreed.

apps/librelingo_audios/librelingo_audios/generate_audios.py
zmbc: Oops, this was a bug. In the refactor, which I'll push soon, I've fixed this and made it clearer what it does.

apps/librelingo_audios/librelingo_audios/generate_audios.py
zmbc: I think just JSON-serializing the array is fine, not sure why I went with a wrapper object. `ensure_ascii=False` allows Unicode characters in the file.

apps/librelingo_yaml_loader/librelingo_yaml_loader/yaml_loader.py
zmbc: Is the issue here that "settings" is used for both the singular and the plural? "Configs" doesn't seem any clearer to me.

If I've understood you correctly, maybe `text_to_speech_settings_list` would be good?

zmbc:
> > > Note that this PR does not include the results of running the script for existing courses.
> > 
> > 
> > why did you decide to not include that? I think it would be worth including that for Spanish for example
> 
> I think this would also be useful because we'd essentially be able to manually integration test that everything just works

Sure, sounds fine to me. But with the recent changes you've made to courses for storing them in different repositories, I'm not sure how to do this, since the `course.yaml` needs to change to add the new settings.

zmbc:
@kantord I think I've responded to most of your comments. The big TODO here for me is adding tests, which I will take a crack at soon.

apps/librelingo_audios/librelingo_audios/generate_audios.py
kantord: I am not sure what the advantages of unicode vs ascii escape sequences are? 🤔 I guess both should work well so I don't mind either way

apps/librelingo_yaml_loader/librelingo_yaml_loader/yaml_loader.py
kantord: > Is the issue here that "settings" is used for both the singular and the plural? 

yeah, I think this was my issue yes 😅 even though maybe I haven't realized it consciously 

> If I've understood you correctly, maybe text_to_speech_settings_list would be good?

sounds perfect to me!

kantord:
> > > > Note that this PR does not include the results of running the script for existing courses.
> > > 
> > > 
> > > why did you decide to not include that? I think it would be worth including that for Spanish for example
> > 
> > 
> > I think this would also be useful because we'd essentially be able to manually integration test that everything just works
> 
> Sure, sounds fine to me. But with the recent changes you've made to courses for storing them in different repositories, I'm not sure how to do this, since the `course.yaml` needs to change to add the new settings.

I just reverted that, because I realized it's breaking. It might take me some time to actually fix it 😅 So if you are fast enough, you might get this merged even before.

If not, then I can manually do the integration test locally, so I guess let's decide on this on the moment, whatever seems most convenient once we're there

kantord:
There's another change @zmbc that I merged. Basically it should allow you to locally test the python packages together without having to publish them or modify pyproject.toml. I'm more or less confident that it should work 😅

Basically the idea is that you just use `poetry` in the root folder. The other packages should be automatically linked in the virtualenv once you run `poetry install`.

Afterwards, `poetry run pytest` should run all tests from all python packages. This always worked, but it was testing the published versions of the packages, instead of the local changes, and it also was very slow (10s+). Now it should be quite fast and in theory it tests local changes, not the published versions.

kantord:
btw, thank you for the progress @zmbc 🚀!!

kantord:
Another thing that's missing is end-user documentation for the new features. Can you please either include documentation, or create a new issue for it? It's probably enough if the issue has a very rough draft/list of things that need to be included in the documentation

zmbc:
> I just reverted that, because I realized it's breaking. It might take me some time to actually fix it. So if you are fast enough, you might get this merged even before.
> 
> If not, then I can manually do the integration test locally, so I guess let's decide on this on the moment, whatever seems most convenient once we're there

... but `/courses/` is still gitignored -- so I still don't have the Spanish course configuration.

zmbc:
> Basically the idea is that you just use poetry in the root folder. The other packages should be automatically linked in the virtualenv once you run poetry install.

I can't get this to work, with either the `pyproject.toml` files in my fork or the originals from your `main` branch. Either way, I get the error `ImportError: cannot import name 'TextToSpeechSettings' from 'librelingo_types'`

apps/librelingo_audios/librelingo_audios/generate_audios.py
zmbc: I'm working on implementing most of these tests. However, it does not behave exactly how you have outlined here. The index JSON file is the *single source of truth* and the files themselves are a write-only side effect -- they are never read to determine state. So A would not be touched, and D and E are exactly the same from the script's perspective.

zmbc:
> Another thing that's missing is end-user documentation for the new features. Can you please either include documentation, or create a new issue for it? It's probably enough if the issue has a very rough draft/list of things that need to be included in the documentation

I've added documentation.

kantord:
> > I just reverted that, because I realized it's breaking. It might take me some time to actually fix it. So if you are fast enough, you might get this merged even before.
> > If not, then I can manually do the integration test locally, so I guess let's decide on this on the moment, whatever seems most convenient once we're there
> 
> ... but `/courses/` is still gitignored -- so I still don't have the Spanish course configuration.

yup, in the meantime I merged it again :see_no_evil: so I'll have to manually do the integration test for you before merge :wink: so feel free to ignore what I said about the Spanish course

apps/librelingo_audios/librelingo_audios/generate_audios.py
kantord: > So A would not be touched

why, if the JSON file tells you which files should exist, and it tells you nothing about A, it kinda follows that A should not exist, no? 

> D and E are exactly the same from the script's perspective

E should be a similar case as A, basically since it's a file that should not have existed in the first place, it should be removed by the same logic that removed A. Afterwards it should be generated like any other non-existent audio

kantord:
> > Basically the idea is that you just use poetry in the root folder. The other packages should be automatically linked in the virtualenv once you run poetry install.
> 
> I can't get this to work, with either the `pyproject.toml` files in my fork or the originals from your `main` branch. Either way, I get the error `ImportError: cannot import name 'TextToSpeechSettings' from 'librelingo_types'`

true, the same happens to me :eyes:

kantord:
> > > Basically the idea is that you just use poetry in the root folder. The other packages should be automatically linked in the virtualenv once you run poetry install.
> > 
> > 
> > I can't get this to work, with either the `pyproject.toml` files in my fork or the originals from your `main` branch. Either way, I get the error `ImportError: cannot import name 'TextToSpeechSettings' from 'librelingo_types'`
> 
> true, the same happens to me eyes

It's because I forgot to put this line into the config:

```toml
librelingo-types = { path = "./apps/librelingo_types", develop = true }
```

after putting it there, I get normal test failures now, not import errors.

I pushed[ this commit to main](https://github.com/kantord/LibreLingo/commit/042d4ad73871306b963d53c6497f959226748cd2), so if you rebase it should work for you as well. (After `poetry install`)

apps/librelingo_audios/librelingo_audios/generate_audios.py
kantord: Also, I was thinking that deleting the unneeded files could be a separate cleanup step, but, after some deliberation I think it really should be built into the fetching command.

This is because there are 2 very good reasons to not have any audio files that are not justified by the metadata json:

- The audios might change a lot when changes are merged to a course. This could create a lot of unused files which would be making deployment unnecessarily inefficient
- We want to avoid having unlicensed audio in the course contents hence any files that are not accounted for should be deleted asap

that being said, if you don't feel like implementing the "cleanup" part in this MR, we can create a new issue for it

zmbc: > why, if the JSON file tells you which files should exist, and it tells you nothing about A, it kinda follows that A should not exist, no?

If we've accepted that nobody should be manually editing the `voice` directory, it will be very unusual (and essentially undefined behavior) if anything besides the script has edited it. And no matter how hard we try, it will *never* be possible to ensure correctness despite edits by other parties -- imagine if someone replaced an audio file with a different file but kept the name the same. So, the added "feature" of handling random changes to the file system will be almost never used, and can never work perfectly anyway.

There is also a considerable benefit to the approach I've taken here, in which the directory state is never read. Keeping two things in sync (phrases in the course, phrases in the index) is a surprisingly deep problem, and you'll see in my code that it is already not simple to do so correctly. Adding a third read-write state (audio files in directory) would add a bunch of code complexity, a lot more tests, more surface area for bugs, etc.

> The audios might change a lot when changes are merged to a course. This could create a lot of unused files which would be making deployment unnecessarily inefficient

These would be deleted, though. If the files were created by the script, they will be in the index file. If they are in the index file, they will be deleted if no longer necessary.

> We want to avoid having unlicensed audio in the course contents hence any files that are not accounted for should be deleted asap

Again, this is undefined behavior/user error -- and on top of that, it would not be possible for this audio to ever be actually played in the course, because if it is not in the metadata file that means it is not referenced in the course either (otherwise it would be overwritten by a newly generated file).

zmbc: > E should be a similar case as A, basically since it's a file that should not have existed in the first place, it should be removed by the same logic that removed A. Afterwards it should be generated like any other non-existent audio

For practical purposes, there is no difference between deleting + re-generating and simply overwriting.

zmbc:
@kantord I am confused by the CI failure -- I cannot reproduce it locally.

apps/librelingo_audios/librelingo_audios/generate_audios.py
kantord: > For practical purposes, there is no difference between deleting + re-generating and simply overwriting.

that's true

kantord:
> @kantord I am confused by the CI failure -- I cannot reproduce it locally.

not sure why it failed, after retrying it now passes :shrug: it must have been some random glitch or network error while running the tests

apps/librelingo_audios/librelingo_audios/generate_audios.py
kantord: > These would be deleted, though. If the files were created by the script, they will be in the index file. If they are in the index file, they will be deleted if no longer necessary.

ok, that makes sense! Looking at the directories then probably doesn't have to be part of this script as your say. _perhaps_ it could be part of a completely separate cleanup script, or not even that.

apps/librelingo_audios/librelingo_audios/update_audios.py
```diff
@@ -0,0 +1,175 @@
...
+    )
+    old_phrase_set = set(
+        [_phrase_identity_info_from_index(p) for p in phrases_with_existing_audios]
+    )
+
+    if settings.destructive:
+        phrases_to_keep = set()
+    else:
+        phrases_to_keep = new_phrase_set & old_phrase_set
+    phrases_to_fetch = new_phrase_set - phrases_to_keep
```
kantord, line 32: using set operations for this is a quite nice way to implement it, I like it!

kantord:
@zmbc I think this should be ready to merge 🎉!

If you agree, then I'll:

- Check out the branch locally
- manually test it with the German course or Spanish (whichever will be easier)
- release the pypi packages
- squash & merge

kantord (approved):
💯

Good job!

Note to self, needs manual testing and release before merge, see: https://github.com/kantord/LibreLingo/pull/1528#issuecomment-908902307

zmbc:
@kantord Yes, this should be good to go.

I haven't changed the `course.yaml` of any courses in this MR, since it seems like you are in the process of moving those out of the repo.

But if you want to change the settings, remember that audio is now disabled by default. So courses with the old `disable audio files` setting can just have that section deleted, while courses without that setting (courses that use audio) will need:

```yaml
Settings:
  Audio:
    Enabled: True
    TTS:
      - Provider: Polly
        Voice: <insert-voice-here>
        Engine: <insert-engine-here>
```

kantord:
> @kantord Yes, this should be good to go.
> 
> I haven't changed the `course.yaml` of any courses in this MR, since it seems like you are in the process of moving those out of the repo.
> 
> But if you want to change the settings, remember that audio is now disabled by default. So courses with the old `disable audio files` setting can just have that section deleted, while courses without that setting (courses that use audio) will need:
> 
> ```yaml
> Settings:
>   Audio:
>     Enabled: True
>     TTS:
>       - Provider: Polly
>         Voice: <insert-voice-here>
>         Engine: <insert-engine-here>
> ```

hi @zmbc , I want to apologize for the delay. This MR deserves to be merged ASAP. I was trying to test it locally a couple days ago but some stuff didn't work properly, but I didn't manage to fix them. Then I fell ill so I didn't make any progress for several days. I'll try to merge this asap from now on.

kantord:
Hi @zmbc! Finally I got the time to actually properly test it! It looks amazing!!!! 💯 🎉

I noticed one small problem that I didn't notice during the code review: it only generates audio for phrases, not for words. I actually misremembered this I think 😅 I was convinced audios only exist for phrases. Anyways, I'll fix this in a follow-up MR 😉

kantord:
@allcontributors please add @zmbc for code!

zmbc:
> it only generates audio for phrases, not for words

I can't believe I put so much thought into this and never realized this :rofl: 

I think I was thinking the same thing (that only phrases have audio). Oops!

kantord:
> > it only generates audio for phrases, not for words
> 
> I can't believe I put so much thought into this and never realized this rofl
> 
> I think I was thinking the same thing (that only phrases have audio). Oops!

Yeah I also can't believe I didn't realize it :-D 

btw amazing job again, it is a truly valuable contribution to this project and it's a really good bases for future work

kantord:
I prepared the implementation for that in this MR: https://github.com/kantord/LibreLingo/pull/1587

kantord:
I'll probably merge it when the pipeline passes, but if you have comments/review I'd still be happy to make changes in a follow up MR. I'm also thinking of doing some refactoring if and when I have time

kantord:
@zmbc I did some refactors https://github.com/kantord/LibreLingo/pull/1591/files

I merged it straight away, but nothing is set in stone, if you'd like you can review it and things can be changed

PR #1610, opened 2021-09-18
ci: verify that test course JSON files are not out of date. 

gdomingu:
## Description

Adds a script and updates semaphore to check if the Course Test Json Files are up to date.

**Is this related to any open issue(s)?**

fixes https://github.com/kantord/LibreLingo/issues/1447

**Is there anything you need help with to get this PR merged?**

I need help resolving CI errors:
```
ImportError: cannot import name 'AudioSettings' from 'librelingo_types' (/home/semaphore/.pyenv/3.8/lib/python3.8/site-packages/librelingo_types/__init__.py)
```

## Checklist

Don't worry if you haven't done everything on this checklist. Simply create your PR, and make sure everything is done later,
or ask for help if you don't know how to do it.

- [ ] The CI is passing (This will need to go in first for CI to pass https://github.com/kantord/LibreLingo/pull/1613)
- [x] I tested my implementation manually
- [ ] I wrote new tests to test my code, or updated existing tests to test my changes

PR #1613, opened 2021-09-19
fix: Enable audio settings test course

gdomingu:
## Description
Enables audio setting on test course. This is so when we run `yarn exportAllCourses` it doesn't result in a diff. 
Prerequisite for https://github.com/kantord/LibreLingo/pull/1610. 

## Checklist

- [x] The CI is passing

### DIFF
--- a/courses/test/course.yaml
+++ b/courses/test/course.yaml
@@ -34,3 +34,7 @@ Course:
 
 Modules:
   - basics
+
+Settings:
+    Audio:
+      Enabled: True


apps/web/src/courses/test/challenges/animals.json
```diff
@@ -10,15 +10,6 @@
       "priority": 0,
       "group": "353b3be9ecc5"
     },
-    {
-      "type": "listeningExercise",
```
gdomingu, line 14: @kantord is this to be expected? The now failing cypress tests, make me wonder. https://kantord.semaphoreci.com/jobs/aae1086d-13d9-48ae-9245-abbec15ca45b

Or do I just need to remove something from the cypress tests?

kantord: this is weird, let me look into this

kantord: I think it's kinda expected. my understanding is that we need to create the audio files that are missing. However, they can be literally empty text files as the export script won't be able to tell, it's only verifying that they exist afaik

kantord: this is because now the exporting scripts should be able to handle missing audio by skipping listening exercises when the audio is missing. Hence it needs to be fetched first, but the test course won't have any way to fetch audio files.

The beautiful solution would be to add a new audio source called "Dummy audio" which will work the same way as the TTS audio source, but instead of fetching stuff form aws, it will just create dummy empty files.

Easy short term solution is to create these dummy files manually.

gdomingu: @kantord thanks for the info. Where would the dummy files need to be added to? 
Also are you saying that if I add the dummy files and rerun `yarn exportAllCourses` will these `"type": "listeningExercise"` will not be deleted?

kantord: >  Where would the dummy files need to be added to?

It would need to be added to this folder:
https://github.com/kantord/LibreLingo/tree/main/apps/web/static/voice

> Also are you saying that if I add the dummy files and rerun yarn exportAllCourses will these "type": "listeningExercise" will not be deleted?

yes

gdomingu: Thank you, I will try to update this today.

gdomingu:
@kantord it turns out I just had to enable a setting to prevent it from being deleted.

kantord:
wow! It's almost accidental that this works  I guess 😹 but it's great for now!!!!

kantord:
Thank you, @gdomingu !

@allcontributors please add @gdomingu for code

PR #1844, opened 2021-12-07
build: remove lockfiles from apps and use only root lockfile

kantord:
those lockfiles should not be necessary as they are only used
to make the environment reproducible. However, in the CI and
other environments we are only using the root poetry lockfile
for that purpose.

See more: https://python-poetry.org/docs/basic-usage/#commit-your-poetrylock-file-to-version-control


kantord:
@szabgab I was thinking about your suggestion and I think in the end these poetry lock files can really be removed. Is this how you imagined it?

PR #1858, opened 2021-12-10
gitignore .yarnrc

szabgab:


kantord:
how did you end up having that file?

szabgab:
I think `yarn set version classic` created it for me. It's content  is

```
# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
# yarn lockfile v1


yarn-path ".yarn/releases/yarn-1.22.17.cjs"
```

szabgab:
If I can remove it then I will do that and then no need to include this in .gitignore, though I don't think this will hurt anyway.

PR #1884, opened 2021-12-23
Addfilenames #1852

### LINKED ISSUE #1852: include yaml filename in Skill, Module, Course objects
In order to be able to report issues in a curse using filenames it would be useful to include the name of the yaml file in the Skill, Modules, and Course objects.
In  `apps/librelingo_types/librelingo_types/data_types.py`

szabgab:
Adding course_dir to the Course object and filename to both the Module and the Skill objects.

fixes #1852

### DIFF
--- a/apps/librelingo_fakes/librelingo_fakes/fakes.py
+++ b/apps/librelingo_fakes/librelingo_fakes/fakes.py
@@ -220,6 +220,7 @@ courseEmpty = Course(
     modules=[],
     dictionary=fake_dictionary,
     repository_url="https://github.com/kantord/LibreLingo/tree/main/courses/spanish-from-english",
+    course_dir="/some/dir",
 )
 
 
@@ -246,6 +247,7 @@ course1 = Course(
             enabled=True, text_to_speech_settings_list=[TextToSpeechSettings()]
         )
     ),
+    course_dir="/some/dir",
 )
 
 course2 = Course(
@@ -258,6 +260,7 @@ course2 = Course(
     ],
     dictionary=fake_dictionary,
     repository_url="https://github.com/kantord/LibreLingo-Spanish-from-English",
+    course_dir="/some/dir",
 )
 
 
--- a/apps/librelingo_fakes/librelingo_fakes/fakes.py
+++ b/apps/librelingo_fakes/librelingo_fakes/fakes.py
@@ -199,7 +199,7 @@ skills = [
     emptySkill,
     Skill(
         name="Mammals and birds",
-        filename="mammals_and_birds.yaml",
+        filename="animals/skills/mammals_and_birds.yaml",
         id=3,
         words=[word1],
         phrases=[],
--- a/apps/librelingo_fakes/librelingo_fakes/fakes.py
+++ b/apps/librelingo_fakes/librelingo_fakes/fakes.py
@@ -117,6 +117,7 @@ word_with_spaces = Word(
 
 emptySkill = Skill(
     name="Neuter",
+    filename="neuter.yaml",
     id=3,
     words=[],
     phrases=[],
@@ -127,6 +128,7 @@ emptySkill = Skill(
 
 skillWithPhrase = Skill(
     name="Masculine",
+    filename="masculine.yaml",
     id=3,
     phrases=[phrase2],
     words=[],
@@ -138,6 +140,7 @@ skillWithPhrase = Skill(
 
 skillWithWord = Skill(
     name="Masculine",
+    filename="masculine.yaml",
     id=3,
     phrases=[],
     words=[word1],
@@ -149,6 +152,7 @@ skillWithWord = Skill(
 
 skillWithPhraseAndWord = Skill(
     name="Masculine",
+    filename="masculine.yaml",
     id=3,
     phrases=[phrase2],
     words=[word1],
@@ -160,6 +164,7 @@ skillWithPhraseAndWord = Skill(
 
 skillWith3Words = Skill(
     name="Masculine",
+    filename="masculine.yaml",
     id=3,
     phrases=[],
     words=[word1, word2, word2],
@@ -170,6 +175,7 @@ skillWith3Words = Skill(
 
 skillWith3Phrases = Skill(
     name="Masculine",
+    filename="masculine.yaml",
     id=3,
     phrases=[phrase1, phrase2, phrase2],
     words=[],
@@ -182,6 +188,7 @@ skills = [
     skillWithPhrase,
     Skill(
         name="Feminine",
+        filename="feminine.yaml",
         id=3,
         words=[word1, word2, word1, word2],
         phrases=[phrase1],
@@ -192,6 +199,7 @@ skills = [
     emptySkill,
     Skill(
         name="Mammals and birds",
+        filename="mammals_and_birds.yaml",
         id=3,
         words=[word1],
         phrases=[],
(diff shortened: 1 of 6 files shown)


apps/librelingo_fakes/librelingo_fakes/fakes.py
```diff
@@ -192,6 +199,7 @@
     emptySkill,
     Skill(
         name="Mammals and birds",
+        filename="mammals_and_birds.yaml",
```
kantord, line 202: should these file names not be relative to the course root, rather than the module folder?

apps/librelingo_types/librelingo_types/data_types.py
```diff
@@ -197,7 +200,7 @@ class Module(
     ### Usage examples:
 
     ```python
-    my_module = Module(title="Basics", skills=[skill1, skill2])
+    my_module = Module(title="Basics", filenama="basic/module.yaml", skills=[skill1, skill2])
```
kantord, line 203:
```suggestion
    my_module = Module(title="Basics", filename="basic/module.yaml", skills=[skill1, skill2])
```

apps/librelingo_yaml_loader/librelingo_yaml_loader/yaml_loader.py
```diff
@@ -316,7 +317,7 @@ def _load_module(path, course):
             'Module file "{}" needs to have module name'.format(filepath)
         )
 
-    return Module(title=title, skills=_load_skills(path, skills, course))
+    return Module(title=title, filename=str(path)[len(course.course_dir)+1:], skills=_load_skills(path, skills, course))
```
kantord, line 320: why `str(path)` instead of just `path`?

apps/librelingo_yaml_loader/librelingo_yaml_loader/yaml_loader.py
```diff
@@ -403,6 +404,7 @@ def load_course(path):
         modules=[],
         settings=None,
         repository_url=course["Repository"],
+        course_dir=str(path),
```
kantord, line 407: why `str(path)` instead of just `path`?

apps/librelingo_yaml_loader/librelingo_yaml_loader/yaml_loader.py
```diff
@@ -268,6 +268,7 @@ def _load_skill(path, course):
 
     return Skill(
         name=name,
+        filename=str(path)[len(course.course_dir)+1:],
```
kantord, line 271: why `str(path)` instead of just `path`?

apps/librelingo_types/librelingo_types/data_types.py
```diff
@@ -134,7 +135,8 @@ class Course(
         ),
         dictionary=[dict_item1, dict_item2, dict_item3, dict_item4],
         repository_url="https://example.com",
-        setttings=Settings()
```
kantord, line 137: wow, what a typo that has been. it's interesting that it never caused any issues

apps/librelingo_yaml_loader/librelingo_yaml_loader/yaml_loader.py
kantord: btw if I understand correctly, this should be done using [os.path.relpath](https://docs.python.org/3/library/os.path.html#os.path.relpath):

```suggestion
        filename=os.path.relpath(path, start=course.course_dir),
```

apps/librelingo_yaml_loader/librelingo_yaml_loader/yaml_loader.py
kantord: https://github.com/kantord/LibreLingo/pull/1884/files#r774698354

apps/librelingo_yaml_loader/librelingo_yaml_loader/yaml_loader.py
kantord: https://github.com/kantord/LibreLingo/pull/1884/files#r774698354

apps/librelingo_fakes/librelingo_fakes/fakes.py
szabgab: yes, though in the fakes it apparently does not matter. I'll update anyway

apps/librelingo_types/librelingo_types/data_types.py
szabgab: ah, so how does it work then? why are the tests passing?

szabgab: Oh, I see you made the typo not I. Wow I was confused here

apps/librelingo_yaml_loader/librelingo_yaml_loader/yaml_loader.py
szabgab: Becasue I get `TypeError: 'PosixPath' object is not subscriptable`

apps/librelingo_yaml_loader/librelingo_yaml_loader/yaml_loader.py
szabgab: yes, that's much better

szabgab:
I think I fixed all of the issues, right?

apps/librelingo_yaml_loader/librelingo_yaml_loader/yaml_loader.py
kantord: ah ok I think I get it. it's a Path type instead of string.

I'm not sure which one is better to use. string is much simpler which is great, on the other hand Path is built in in python as well. so it could also be useful to provide that type here as well and rather convert it when necessary to display and error message or such

apps/librelingo_fakes/librelingo_fakes/fakes.py
kantord: yeah in the fakes it won't matter right now. I think it's better just for how it will appear in a text failure for example. If someone sees a test failure and they see the path does not look normal they might be distracted by that

apps/librelingo_types/librelingo_types/data_types.py
kantord: yes the typo is just in the string, it's not actual code that's compiled. it would be nice to have some linting tool that can notice these mistakes even in documentation. 

some time ago I was experimenting with the idea of forcing code in documentatino to compile. Not sure if there's a way to do that in python :thinking: 

perhaps doctest would be that actually https://docs.python.org/3/library/doctest.html

but anyways that goes beyond the scope of this PR

apps/librelingo_yaml_loader/librelingo_yaml_loader/yaml_loader.py
kantord: with relpath this hopefully won't happen, right? I actually never used that in Python though

kantord:
> I think I fixed all of the issues, right?

2021-12-23T20:18:12+01:00

I can see some inconsistency now how the filenames are handled in the fakes

kantord:
2021-12-23T20:19:54+01:00

also some tests are failing now, please check out the CI output https://kantord.semaphoreci.com/workflows/8c739b5a-790a-4ec9-b5c4-49d38cfdbfbb?pipeline_id=5c82391e-6e48-4186-8be5-75f3b2e921d4

kantord:
code style (black) is still failing and also the readme files need to be updated for some reason :-) 

please check out the failures, I believe the failures have instructions how to fix them

szabgab:
Oh I though they were related to Node JS as that what the is listed in the report. I clicked through to the actual reports, but they are empty. Anyway I could run them locally and fix them. Let's see what does the CI say now.

kantord:
thanks @szabgab !!

@allcontributors please add @szabgab for code

Issue #1889, opened 2021-12-24
use doctest to test the code-snippets in the documentation


szabgab:
In the python code, that's it

PR #1894, opened 2021-12-24
Convert text documentation to executable documentation for #1889

szabgab:
progress on #1889

There are a few more remaining cases, but they need some setup so I am not sure they should be converted.

### DIFF
--- a/apps/librelingo_types/README.md
+++ b/apps/librelingo_types/README.md
@@ -26,13 +26,12 @@ Settings about how to use TTS to generate audios
 
 ### Usage example:
 
-```python
-TextToSpeechSettings(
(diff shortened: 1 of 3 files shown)


kantord:
> but they need some setup so I am not sure they should be converted.

I would say so. If they are so hard to convert then probably they should be changed or simplified, or the overall API should be simplified.

Can you show me some examples of such cases?

szabgab:
The Course, Module, and Skill object all need some other object in their initialization. So I have not converted them yet.
in https://github.com/kantord/LibreLingo/blob/main/apps/librelingo_types/librelingo_types/data_types.py

PR #1904, opened 2021-12-25
fix bare-except  #1899

### LINKED ISSUE #1899: pylint config: enable "bare-except"

**How to solve**

**Why solve like this**

**Additional context**

szabgab:
fixes #1899

### DIFF
--- a/apps/librelingo_yaml_loader/librelingo_yaml_loader/yaml_loader.py
+++ b/apps/librelingo_yaml_loader/librelingo_yaml_loader/yaml_loader.py
@@ -227,11 +227,14 @@ def _sanitize_markdown(mdtext):
 
 def _load_introduction(path):
     "Loads the introduction text from a Markdown file"
+    if not os.path.exists(path):
+        return None
+
     try:
         with open(path) as f:
             return _sanitize_markdown(f.read())
-    except:
-        return None
+    except Exception as err:
+        raise RuntimeError(f'Makrdown file "{path}" could not be loaded: {err}')
 
 
 def _load_skill(path, course):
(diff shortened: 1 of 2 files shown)


apps/librelingo_yaml_loader/librelingo_yaml_loader/yaml_loader.py
```diff
@@ -227,11 +227,14 @@ def _sanitize_markdown(mdtext):
...
+    if not os.path.exists(path):
+        return None
+
     try:
         with open(path) as f:
             return _sanitize_markdown(f.read())
-    except:
-        return None
+    except Exception as err:
+        raise RuntimeError(f'Makrdown file "{path}" could not be loaded: {err}')
```
kantord, line 237: hmm, in this case perhaps the whole try-except could be removed altogether?

because if the file does not exist, it's already returning `None`. In any other case it should actually probably fail with the exception, no?

szabgab:
That was my first thought, but then I saw the code below where the exceptions are caught and the re-raised with some explanation. In any case if you'd prefer this code to be without try-except I can remove it.

kantord:
> That was my first thought, but then I saw the code below where the exceptions are caught and the re-raised with some explanation. In any case if you'd prefer this code to be without try-except I can remove it.

but in this case it's different because the markdown file is optional

kantord:
thanks @szabgab 

btw I am rewriting your commit messages when I merge because they don't match the standard format (hence the CI check failures calling attention to this fact as well)

see this page for the documentation regarding conventional commit messages https://www.conventionalcommits.org/en/v1.0.0/

Issue #1803, opened 2021-11-26
Move all the courses to separate repositories


szabgab:
There are a few in `temporarily_inactive_courses/` and there is `courses/basque-from-english/` that would be better in separate repositories.

It might be a good idea to set up a GitHub organization for this (e.g. LibreLingo ) and move all the repositories there.

kantord:
> It might be a good idea to set up a GitHub organization for this (e.g. LibreLingo ) and move all the repositories there.

perhaps. But a lot of people want to host courses outside of GitHub already. 

Probably best to start creating a page in the GUI (as opposed to simply the documentation) for showing the different course repos. Initially this could lead to the web gui of the git repo (wherever it might be hosted) and later we can use it as an entry point for the course editor perhaps. (whenever we get to that point)

frostedkitsune:
Yes, I would also go for a seperate single source of course and GUI, and that would be only possible if we have our own course management system, a gui course editor with yaml template, git based hosting.....enabling ppl to use raw git repo seems kinda messy to me

frostedkitsune:
> It might be a good idea to set up a GitHub organization for this (e.g. LibreLingo ) and move all the repositories there.
Indeed, but doing that would cause some problems as the repo url will be changed and the previous url probably wouldn't be accessiable leaving the contributors  in no where , and developement workflows would be broken for few days, then all the articles available online reffering to repo url would also be broken, I have done this before on gitlab, dunno how github handles this, so these things might happen.

another less catastrophic approach would be creating an org then forking this repo under that org and then we would have to notify the contributors to visit new repo url though showing messages in this repo's readme.

kantord:
> .enabling ppl to use raw git repo seems kinda messy to me.

I wouldn't say it's messy, when it's only technical people  who will use it that way. But we need to be able to abstract it away for people who are not technical. An example of achieving this is https://www.gitbook.com/

frostedkitsune:
> But we need to be able to abstract it away for people who are not technical. 

yes, that's what I want to say, ppl who would create course might not be much technical to handle the unnecessary complications of git and github, neither they should! Can't we create a separate git repo for courses and add it to this repo as a git submodule? That would be nice, no?

szabgab:
There is no need to use submodules. We can maintain a file (we already have one) with the list of languages that need to be processed and their source) and the deployment process can use that.

szabgab:
All the courses are now in separate repos

PR #1908, opened 2021-12-25
make ids unique

szabgab:


kantord:
this is a bit wrong. all of those IDs should actually be UUIDs in my opinion. otherwise it will be very annoying to keep them unique. or near impossible if we are considering that different deployments of different librelingo instances can exist in different places and we can't definitively enforce the uniqueness of these values across github repositories.

Perhaps it could even be a good idea to create an API to register librelingo skills and returns ids from there :thinking:  then skills can be registered in one place and there can be a database of globally unique skill IDs :thinking:  but maybe it's overkill and I'm not sure if even this could be used to effectively enforce the uniqueness globally, as someone might inadvertently copy and id from another repository

szabgab:
Maybe they should not be manually maintained, but they are now. I have a script in the https://github.com/szabgab/LibreLingo-tools repo that checks for duplicate ids. I run it as the CI of the courses I write and the dev view I started to create.
That's how I found these. I think, for now, instead of trying to come up with an ultimate solution with registration an all,  it would be easier to accept this PR and add the Id checking to the CI of all the courses.
(But first I'd first to #1803)

PR #1937, opened 2021-12-31
Qbecb1zen/pylint simplifiable if expression

### LINKED ISSUE #1927: pylint config: enable "simplifiable-if-expression"

**How to solve**

**Why solve like this**

**Additional context**

qbecb1zen:
# edit this template according to your needs, delete anything that's not relevant

## Description

_Short description of the changes you made. Not necessary if it's clear based on the issue you're resolving_

fixes #1927

Last check isn't needed since it's just a style PR
_..._

## Checklist

- [x] The CI is passing

### DIFF
--- a/poetry.toml
+++ b/poetry.toml
@@ -1,0 +1,4 @@
+# enforce the local creation of .venv
+[virtualenvs]
+create = true
+in-project = true
--- a/poetry.toml
+++ b/poetry.toml
@@ -1,4 +1,0 @@
-# enforce the local creation of .venv
-[virtualenvs]
-create = true
-in-project = true
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -731,6 +731,24 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login": "qbecb1zen",
+      "name": "Amine Louzar",
+      "avatar_url": "https://avatars.githubusercontent.com/u/66479002?v=4",
+      "profile": "https://github.com/qbecb1zen",
+      "contributions": [
+        "code"
+      ]
+    },
+    {
+      "login": "najmieh",
+      "name": "Najmieh Sadat Safarabadi",
+      "avatar_url": "https://avatars.githubusercontent.com/u/31998054?v=4",
+      "profile": "https://github.com/najmieh",
+      "contributions": [
+        "code"
+      ]
     }
   ],
   "contributorsPerLine": 7,
--- a/README.md
+++ b/README.md
@@ -184,7 +184,7 @@ The mascot is designed by [@almostdesigner](https://github.com/almostdesigner).
 
 ### Contributors
 <!-- ALL-CONTRIBUTORS-BADGE:START - Do not remove or modify this section -->
-[![All Contributors](https://img.shields.io/badge/all_contributors-77-orange.svg?style=flat-square)](#contributors-)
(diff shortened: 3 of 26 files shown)


poetry.toml
```diff
@@ -0,0 +1,4 @@
+# enforce the local creation of .venv
+[virtualenvs]
+create = true
+in-project = true
```
qbecb1zen, line 4: To avoid trailing .venv on developers machine and to have the .venv directly on the project (as it is the case with node_modules for example).

Just a suggestion related to my ex projects with poetry. If you think it's not a good idea, i can remove it 😇

kantord: > To avoid trailing .venv on developers machine and to have the .venv directly on the project (as it is the case with node_modules for example).

hmm, is it a good idea? :thinking: 

I thought it was kinda convenient that the virtualenv is not there, like node_modules :thinking: 

btw actually yarn2 does the same as poetry by default and not put node modules in the project folder :thinking: 

> To avoid trailing .venv on developers machine

but this part is interesting, do you mean poetry creates a lot of dead virtual envs or such? :thinking:

szabgab: I thought we are going in the direction of moving also the `node_modules` out of the local directory.

qbecb1zen: Ohh, and what is convenient in having all the virtualenvs/node_modules no longer inside the projet itself ?

For me it's the same thing except that when having them inside the local directory of the project, you can have some useful commands such as hard-install (to delete the `.venv` and add a new one).
And also honestly, just as a  lambda developer/contributor, it may be frustrating to have a myriad of trailing `.venv` in your local machine just because you've run a `poetry install` once in a project and then you deleted the project, but the `.venv` created for this project is still persistant in the machine

Also i think there is a big difference in managing dependencies between poetry and yarn2, since yarn2 create this .pnp.cjs to map dependencies on the disk, while poetry just creates a virtual environment everytime, so if you want to have a fresh new venv, you have to do it manually.

This problem isn't the same in node, because of the dependency tree management. I think that the best way to make analogy between python and node for packages management is by following the [pep 582 suggestion](https://www.python.org/dev/peps/pep-0582/), which is currently in development by frostming in this new package manager that will act more like npm/yarn than poetry/pipenv https://pdm.fming.dev/

All that to say, that to me it's mainly the same, it was just a suggestion, and i wanted to have some feedbacks to build an opinion on my way of doing things on other projects (where i opted for this option for example) so pretty eager to know what  are the reasons to not have the node_modules/.venv in the repo of the project! 😇🚀

~Qbecb1zen

kantord: > Also i think there is a big difference in managing dependencies between poetry and yarn2, since yarn2 create this .pnp.cjs to map dependencies on the disk, while poetry just creates a virtual environment everytime, so if you want to have a fresh new venv, you have to do it manually.

that's a very good point!

kantord: I think we can try this, I'm not too bothered by having the virtualenv in the project folder I guess. I actually see that it's a bit easier to see where the virtualenv is and it can be nice to somehow be able to easily delete it :thinking: 

unless perhaps poetry already has some built in way to manage this better. similar to yarn2 maybe

qbecb1zen: I will remove this file and maybe it's a subject to discuss later, but there is already a way to delete the virtualenv i just discovered quickly using the poetry cli! i am making this PR ready for review asap :) 

Screenshot 2021-12-31 at 22 23 22

kantord: ok! thank you @qbecb1zen 

btw good job on your pull requests so far

PR #1950, opened 2022-01-02
refactor(tests): refactor test_export.py

qbecb1zen:
# edit this template according to your needs, delete anything that's not relevant

## Description

refactoring the test_export file into more modular one for more modular tests and more readable tests

fixes #{issue id here}

_..._

## Checklist

- [ ] The CI is passing
- [ ] I tested my implementation manually


qbecb1zen:
@kantord  just a question, is those security checks important for the tests? (it's also bypassable whenever we don't use snake_case )

Screenshot 2022-01-02 at 22 08 17

apps/librelingo_json_export/tests/test_challenges_clean_word.py
```diff
@@ -0,0 +1,30 @@
+from utils import FakeLanguages
+from librelingo_utils import clean_word
+
+
+def test_empty_string():
+    assert clean_word("") == ""
+
+
+def test_removes_parentheses():
+    assert clean_word("(foo") == FakeLanguages.LANG_1
```
kantord, line 10: this is confusing

1. why extract this value to a constant/class? I don't see the point in that as this function basically works on generic strings. 
2. `FakeLanguages.LANG_1` sounds like this is a language. But it's actually a cleaned word.

I would probably remove this `FakeLanguages` thing

qbecb1zen: Oh you are right, I am still trying to adapt 100% with the codebase, the tests and all the workspace of the current work, i created the enum for the FakeWords "foo" "bar" as there were a lot of repetition of them just to void this string repetition (i will probably have to change the `clean_word("foo")` into a `clean_word(f"({...}")`, i missed some of the `foo` used. But why do you believe it's a bad idea to have an enum to avoid code repetition since it allow us if once we want to change all the "foo" into "foozibarzibar" to do it just once instead of having to recheck all over the codebase to find them ?

I also was confused a little bit why it uses constants words as "foo" but also not constant ints (with randint) since the tests are normally made to have exactly the same behavior even if we run them 10000 times to have a deterministic behavior with values.
While the random values may lead to a failing tests due to a edge case value that will not be reproducable or it will but with a 1/5000 probability.

qbecb1zen: a discussion i was inspired of when i first started working on testing:

https://softwareengineering.stackexchange.com/questions/221632/testing-deterministic-or-non-deterministic#:~:text=When-,every,-run%20of%20the

https://stackoverflow.com/questions/32458/random-data-in-unit-tests

Also the "foo" "bar" ... values maybe can be more "realistic" words that can appear in the real app?

qbecb1zen: Ok i rollbacked it to "foo" "bar" variables, but just poining out that this is what misleaded me :) 

```python
        x, y = "foo", "bar"
        z = str(fakes.fake_value())
        p1 = fakes.customize(fakes.word1, in_source_language=[x])
        p2 = fakes.customize(fakes.word1, in_source_language=[x, y])
        self.assertEqual(get_dumb_opaque_id(z, p1), get_dumb_opaque_id(z, p2))
```

Would like to have your input concerning the actual naming and repetition on variables!

Screenshot 2022-01-07 at 04 16 02

kantord:
> @kantord just a question, is those security checks important for the tests? (it's also bypassable whenever we don't use snake_case )
> 
> Screenshot 2022-01-02 at 22 08 17

this can be bypassed. the warning makes sense but it is not concerning in our case because we are not using these numbers for any production use case, just for testing

apps/librelingo_json_export/tests/test_challenges_clean_word.py
kantord: > created the enum for the FakeWords "foo" "bar" as there were a lot of repetition of them just to void this string repetition

is that type of string repetition bad? 

the main reason I prefer it this way is because this way it demonstrates the usage of the function more (no need to look up the meaning of the constant)

kantord: > I also was confused a little bit why it uses constants words as "foo" but also not constant ints (with randint) since the tests are normally made to have exactly the same behavior even if we run them 10000 times to have a deterministic behavior with values.
While the random values may lead to a failing tests due to a edge case value that will not be reproducable or it will but with a 1/5000 probability.

this is kind of a good point, using this type of randomness is probably not the best idea

kantord: > Also the "foo" "bar" ... values maybe can be more "realistic" words that can appear in the real app?

yes, and this would effectively be a better way to also get rid of the code duplication

kantord (approved):

kantord (changes requested):
one small change as indicated in the comments. regarding changing "foo" and the like, I'll leave it up to you whether or not you want to do that!

qbecb1zen:
> one small change as indicated in the comments. regarding changing "foo" and the like, I'll leave it up to you whether or not you want to do that!

i will change the random, but for the "foo" and new phrases, maybe a new issue/PR is better so if anyone else is interested on it he can jump on it (it's also easier for review than having it included in this 500 changes pr), otherway whenever i have a moment to have a deeper look in the codebase (as i am doing now with easy tasks like refactoring) to have a better idea on the architecture etc :) !

kantord:
> but for the "foo" and new phrases, maybe a new issue/PR is better so if anyone else is interested on it he can jump on it (it's also easier for review than having it included in this 500 changes pr), otherway whenever i have a moment to have a deeper look in the codebase (as i am doing now with easy tasks like refactoring) to have a better idea on the architecture etc :) !

sure

kantord:
> (personally i am taking time to read all the code base and as i don't have too much time, i feel like it can be improved on this side to offer a easier onboarding for everyone that would like to contribute)

for sure.

The work that you've already done and what you are doing is really valuable. We really need to make the code base as approachable as possible for new contributors. btw what do you think about the development documentation? https://librelingo.app/docs/

some people have complained that it's not the best and it's not that easy to get started. if you have any ideas how to improve it, please create some issues :-)

kantord:
I merge it as is now, but feel free to open some follow up PRs if there's something you are not satisfied with!

PR #1955, opened 2022-01-04
refactor(tests): add testing_utilities.py

vil02:
Continuation of [rewriting all of the tests using pytest](https://github.com/kantord/LibreLingo/pull/1912#issuecomment-1002282919).

- ~I added the file [`testing_utilities.py`](https://github.com/vil02/LibreLingo/blob/aa3c027f55b6b54e43a7451a3ca2fa67a5aabc8c/apps/librelingo_yaml_loader/tests/testing_utilities.py), with two functions [`get_some_int()`](https://github.com/vil02/LibreLingo/blob/aa3c027f55b6b54e43a7451a3ca2fa67a5aabc8c/apps/librelingo_yaml_loader/tests/testing_utilities.py#L6) and [`get_some_str()`](https://github.com/vil02/LibreLingo/blob/aa3c027f55b6b54e43a7451a3ca2fa67a5aabc8c/apps/librelingo_yaml_loader/tests/testing_utilities.py#L14). They allow to avoid the _duplicated_ pieces of code like [`random.randint(0, 1000)`](https://github.com/kantord/LibreLingo/blob/1d33df8ecb9bc52e63a220f33fdbdaa7964d44b8/apps/librelingo_yaml_loader/tests/test_yaml_loader.py#L43) or [`str(random.randint(0, 1000))`](https://github.com/kantord/LibreLingo/blob/1d33df8ecb9bc52e63a220f33fdbdaa7964d44b8/apps/librelingo_yaml_loader/tests/test_yaml_loader.py#L48) respectively. Because mentioned functions do not need to be cryptographically-secure, I have suppressed the [sonar warning](https://github.com/vil02/LibreLingo/blob/aa3c027f55b6b54e43a7451a3ca2fa67a5aabc8c/apps/librelingo_yaml_loader/tests/testing_utilities.py#L11)~ In order to avoid _duplicated_ pieces of code like [`random.randint(0, 1000)`](https://github.com/kantord/LibreLingo/blob/1d33df8ecb9bc52e63a220f33fdbdaa7964d44b8/apps/librelingo_yaml_loader/tests/test_yaml_loader.py#L43) or [`str(random.randint(0, 1000))`](https://github.com/kantord/LibreLingo/blob/1d33df8ecb9bc52e63a220f33fdbdaa7964d44b8/apps/librelingo_yaml_loader/tests/test_yaml_loader.py#L48) I followed @kantord's [suggestion](https://github.com/kantord/LibreLingo/pull/1955#discussion_r779361178) and applied functions from [`fakes.py`](https://github.com/kantord/LibreLingo/blob/main/apps/librelingo_fakes/librelingo_fakes/fakes.py),
- I moved the methods of the class [`TestConvertPhrase`](https://github.com/kantord/LibreLingo/blob/dac6ef05e72984e04645ff0fd4e59cb1372cf0fc/apps/librelingo_yaml_loader/tests/test_yaml_loader.py#L820) and the function [`test_convert_phrase_complains_about_missing_translation()`](https://github.com/kantord/LibreLingo/blob/dac6ef05e72984e04645ff0fd4e59cb1372cf0fc/apps/librelingo_yaml_loader/tests/test_yaml_loader.py#L1119) to a new file [`test_yaml_loader_convert_phrase.py`](https://github.com/vil02/LibreLingo/blob/bc4af38b0ad501f308df745a158e75e1e01a195c/apps/librelingo_yaml_loader/tests/test_yaml_loader_convert_phrase.py),
- I moved the methods of the class [`TestConvertPhrases`](https://github.com/kantord/LibreLingo/blob/dac6ef05e72984e04645ff0fd4e59cb1372cf0fc/apps/librelingo_yaml_loader/tests/test_yaml_loader.py#L796) to a new file [`test_yaml_loader_convert_phrases.py`](https://github.com/vil02/LibreLingo/blob/a5243805a9e740a158a881beac03e4670d3ac45a/apps/librelingo_yaml_loader/tests/test_yaml_loader_convert_phrases.py),
- I moved the methods of the class [`TestConvertWords`](https://github.com/kantord/LibreLingo/blob/dac6ef05e72984e04645ff0fd4e59cb1372cf0fc/apps/librelingo_yaml_loader/tests/test_yaml_loader.py#L716) to a new file [`test_yaml_loader_convert_words.py`](https://github.com/vil02/LibreLingo/blob/0c2e9e519d958e7865084ce1039bf27469e7dc0c/apps/librelingo_yaml_loader/tests/test_yaml_loader_convert_words.py),
- I moved the methods of the class [`TestConvertWord`](https://github.com/kantord/LibreLingo/blob/dac6ef05e72984e04645ff0fd4e59cb1372cf0fc/apps/librelingo_yaml_loader/tests/test_yaml_loader.py#L740) to a new file [`test_yaml_loader_convert_word.py`](https://github.com/vil02/LibreLingo/blob/e7c09953d00ab8a8487710bafa7677a963cfd684/apps/librelingo_yaml_loader/tests/test_yaml_loader_convert_word.py).


apps/librelingo_yaml_loader/tests/testing_utilities.py
```diff
@@ -0,0 +1,19 @@
...
+    """
+    Returns some 'aleatory' non-negatve integer.
+    It is NOT cryptographically-secure.
+    """
+    return random.randint(0, 1000)  # NOSONAR
+
+
+def get_some_str():
+    """
+    Returns some 'aleatory' string.
```
kantord, line 16:
```suggestion
    Returns a random string.
```

apps/librelingo_yaml_loader/tests/testing_utilities.py
```diff
@@ -0,0 +1,19 @@
...
+
+def get_some_int():
+    """
+    Returns some 'aleatory' non-negatve integer.
+    It is NOT cryptographically-secure.
+    """
+    return random.randint(0, 1000)  # NOSONAR
+
+
+def get_some_str():
```
kantord, line 14:
```suggestion
def get_random_str():
```

apps/librelingo_yaml_loader/tests/testing_utilities.py
```diff
@@ -0,0 +1,19 @@
+"""Contains utilities for tests of the librelingo_yaml_loader"""
+
+import random
+
+
+def get_some_int():
+    """
+    Returns some 'aleatory' non-negatve integer.
```
kantord, line 8:
```suggestion
    Returns some random non-negatve integer.
```

apps/librelingo_yaml_loader/tests/testing_utilities.py
```diff
@@ -0,0 +1,19 @@
+"""Contains utilities for tests of the librelingo_yaml_loader"""
+
+import random
+
+
+def get_some_int():
```
kantord, line 6:
```suggestion
def get_random_int():
```

kantord: btw there's already a package called librelingo fakes. And I think it already has these generators :thinking:

kantord: at least, there's something similar: https://github.com/kantord/LibreLingo/blob/main/apps/librelingo_fakes/librelingo_fakes/fakes.py#L270

apps/librelingo_yaml_loader/tests/test_yaml_loader.py
```diff
@@ -40,12 +35,12 @@ def create_patch(self, module):
         patcher = patch(module)
         self.addCleanup(patcher.stop)
         mock = patcher.start()
-        mock.return_value = random.randint(0, 1000)
+        mock.return_value = fakes.fake_value().id
```
kantord, line 38: why `.id`?

apps/librelingo_yaml_loader/tests/testing_utilities.py
vil02: In few places the `random_path` used to be a [`string`](https://github.com/kantord/LibreLingo/blob/37a04d004cbc341f0ffe8b256c1f5a6a168539ad/apps/librelingo_yaml_loader/tests/test_yaml_loader.py#L986) and now it is a proper [`pathlib.Path`](https://github.com/vil02/LibreLingo/blob/1bfccff9fc72d5ea1e1da3a0d2f3199283fea038/apps/librelingo_yaml_loader/tests/test_yaml_loader.py#L828). In my opinion it was meant to be like this, but please have a look.

apps/librelingo_yaml_loader/tests/test_yaml_loader.py
vil02: Before `mock.return_value` was an `int` not a `namedtuple("FakeValue", ["id"])`.

vil02: What do you think about adding a function `number()` to [`fakes.py`](https://github.com/kantord/LibreLingo/blob/37a04d004cbc341f0ffe8b256c1f5a6a168539ad/apps/librelingo_fakes/librelingo_fakes/fakes.py) and
- apply it in [`fake_value()`](https://github.com/kantord/LibreLingo/blob/37a04d004cbc341f0ffe8b256c1f5a6a168539ad/apps/librelingo_fakes/librelingo_fakes/fakes.py#L270),
- use it instead of `fakes.fake_value().id`.

kantord: yes this makes a lot of sense, @vil02 !!

apps/librelingo_yaml_loader/tests/testing_utilities.py
kantord: I am not sure what you mean, what does the path thing have to do with string? :thinking:  sorry I'm a bit confused.

in any case I don't mean that you have to use that exact one. You can just add a new fake in that library, I think it would be a good idea to be able to share it between all packages.

vil02: In the old code, the variable `randomPath`
https://github.com/kantord/LibreLingo/blob/37a04d004cbc341f0ffe8b256c1f5a6a168539ad/apps/librelingo_yaml_loader/tests/test_yaml_loader.py#L986
was a `string` like `'123'`.
Now `random_path`
https://github.com/vil02/LibreLingo/blob/1bfccff9fc72d5ea1e1da3a0d2f3199283fea038/apps/librelingo_yaml_loader/tests/test_yaml_loader.py#L828
is a `pathlib.Path` like `pathlib.Path('./123')`.
I think the current situation makes more sense: the variable was `randomPath` not `randomFolderName`. And the behaviour is the same: tested functions converts the input into `pathlib.Path` anyway.

apps/librelingo_fakes/README.md
```diff
@@ -31,3 +31,13 @@ fake_course = fakes.customize(fakes.course1, modules=[
...
 ```
 
+<a name="librelingo_fakes.fakes.number"></a>
+#### number
+
+```python
+number(max_value=10000)
+```
+
+Returns some 'random' non-negatve integer.
```
qbecb1zen, line 41:
```suggestion
Returns a 'random' non-negatve integer.
```

qbecb1zen: Also btw, regarding to this discussion, https://github.com/kantord/LibreLingo/pull/1950#discussion_r777913090
Isn't it better to decide on the edge case or just regular case if there is no edge case, to test only this case instead of having random values ?

vil02: > Also btw, regarding to this discussion, [#1950 (comment)](https://github.com/kantord/LibreLingo/pull/1950#discussion_r777913090) Isn't it better to decide on the edge case or just regular case if there is no edge case, to test only this case instead of having random values ?

This is beyond the scope of this pull request.

kantord: I agree with both @qbecb1zen and @vil02

apps/librelingo_yaml_loader/tests/testing_utilities.py
kantord: ok, that makes sense

apps/librelingo_yaml_loader/tests/test_yaml_loader_convert_phrase.py
```diff
@@ -0,0 +1,82 @@
...
+from librelingo_yaml_loader.yaml_loader import _convert_phrase
+
+from librelingo_fakes import fakes
+
+
+def _alt_versions():
+    return "Alternative versions"
+
+
+def _alt_translations():
```
kantord, line 17: what's the purpose of functions like these? to avoid string repetition?

I don't think it makes sense like this :thinking:

apps/librelingo_yaml_loader/tests/test_yaml_loader_convert_phrase.py
```diff
@@ -0,0 +1,82 @@
...
+def test_includes_main_version(raw_fake_phrase):
+    assert (
+        _convert_phrase(raw_fake_phrase).in_target_language[0]
+        == raw_fake_phrase["Phrase"]
+    )
+
+
+def test_includes_alternative_versions(raw_fake_phrase):
+    result = _convert_phrase(raw_fake_phrase).in_target_language
+    assert raw_fake_phrase[_alt_versions()][0] in result
```
kantord, line 51: this is a weird usage, why not just use the string :thinking: or is there some anti-duplication rule that enforces this (pylint?) :thinking:

vil02: Sonar was complaining about it (some "code smell").

apps/librelingo_yaml_loader/tests/test_yaml_loader_convert_phrase.py
vil02: Cf. [my comment](https://github.com/kantord/LibreLingo/pull/1955#discussion_r780655919). I think that is a good way to have a "read-only constant". Of course, if you will try hard enough, you can overwrite everything, but now you will at least have a warning.

apps/librelingo_yaml_loader/tests/test_yaml_loader_convert_phrase.py
kantord: why is it a code smell and is there a different way to deal with it?

apps/librelingo_yaml_loader/tests/test_yaml_loader_convert_phrase.py
kantord: but I think it does not need to be a constant :thinking:

apps/librelingo_yaml_loader/tests/test_yaml_loader_convert_phrase.py
vil02: The complain was the string `"Alternative versions"` was repeated more then 3 (?, I am not sure) times. Similarly with `"Alternative translations"`.

Why is this a code smell? I think it is quite clear why to [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself). In this specific case I would also argue that avoiding duplication reduces the probability of making a typo in future tests.

Is there a different way? Well:
- simply define a _constant_ `_ALT_VERSIONS = "Alternative versions"` (but people are more keen to overwrite a string than a function),
- define a function 
```python
def get_alternative_versions(in_raw_data):
    return in_raw_data["Alternative versions"]
```
and use it when needed. One could argue here that such approach separates the test from the structure of the raw data. But in principle one should also write tests for the function `get_alternative_versions()`.

Do you know another way (besides removing few of the problematic tests)?

apps/librelingo_yaml_loader/tests/test_yaml_loader_convert_phrase.py
vil02: This variable is only being read. So it should be a constant.

kantord: It does not even need to be a variable. It's just a key of a dictionary in a JSON/YAML structure 🤔 

this should be more or less equivalent to a simple dot notation, but that's not really available for Python and would not work well anyway since some of the keys have spaces 🤔

I don't see this really as a constant or variable though. Or can you show me some example or argument why use this notation for these purposes?

I think if we are feeling really bad about using string literals for getting values of dicts then probably the alternative would be to create a class for each of these dictionary types and using objects of those class. But that also sounds like unnecessary complication just to get a value from a dict 🤔

Or else if you really don't want string literals, why not do this?

```python
def get_alternative_translations(phrase: YamlPhrase):
  return phrase["Alternative translations"]
```

apps/librelingo_yaml_loader/tests/test_yaml_loader_convert_phrase.py
kantord: > simply define a constant _ALT_VERSIONS = "Alternative versions" (but people are more keen to overwrite a string than a function),

actually I don't think this reduces the chance of a typo. you will get a different error but in most examples you should get an error nevertheless 🤔 

I can see the "define a function" idea that you are citing is the same I just wrote! I think that's a good idea.

> Do you know another way (besides removing few of the problematic tests)?

Perhaps instead of random fake  values you can use some concrete example, then

```python
assert raw_fake_phrase[_alt_versions()][0] in result
```

maybe be replaced with

```python
assert "She sells sea shells" in result
```

kantord (approved):

kantord:
Thanks!

PR #1972, opened 2022-01-10
refactor(tests): use @pytest.mark.parametrize

vil02:
## Description

I parametrized some of the tests.

### DIFF
--- a/apps/librelingo_json_export/tests/test_challenges_clean_word.py
+++ b/apps/librelingo_json_export/tests/test_challenges_clean_word.py
@@ -1,29 +1,27 @@
+import collections
+import pytest
+
 from librelingo_utils import clean_word
 
-
-def test_empty_string():
-    assert clean_word("") == ""
+TCData = collections.namedtuple("TCData", ["input_str", "expected", "name"])
 
 
-def test_removes_parentheses():
-    assert clean_word("(Hello there") == "Hello there"
-
-
-def test_removes_comma():
-    assert clean_word("This is,") == "This is"
-
-
-def test_doesnt_remove_parts_of_word_1():
-    assert clean_word("Mother-in-law") == "Mother-in-law"
-
-
-def test_doesnt_remove_parts_of_word_2():
-    assert clean_word("It's him") == "It's him"
-
-
-def test_removes_exclamation_mark():
-    assert clean_word("It's his mother-in-law!") == "It's his mother-in-law"
-
-
-def test_weird_english_posessive():
-    assert clean_word("cats'") == "cats'"
+@pytest.mark.parametrize(
+    "test_case_data",
+    [
+        TCData("", "", "empty_string"),
+        TCData("(Hello there", "Hello there", "removes_parentheses"),
+        TCData("This is,", "This is", "removes_comma"),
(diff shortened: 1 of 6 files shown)


apps/librelingo_json_export/tests/test_challenges_clean_word.py
```diff
@@ -1,29 +1,27 @@
...
-def test_removes_exclamation_mark():
-    assert clean_word("It's his mother-in-law!") == "It's his mother-in-law"
+import collections
+import pytest
 
+from librelingo_utils import clean_word
 
-def test_weird_english_posessive():
-    assert clean_word("cats'") == "cats'"
+TCData = collections.namedtuple("TCData", ["input_str", "expected", "name"])
```
kantord, line 6: hmm it's weird for me to create a namedtuple just for this. though I definitely see the value in having fixed labels/names for each field instead of destructuring them "blindly" 

that being said it still feels weird. 

have you taken this idea from somewhere? :thinking:

kantord: in any case, if we keep it maybe I'd prefer the name 

```suggestion
TestCaseData = collections.namedtuple("TestCaseData", ["input_str", "expected", "name"])
```

or 

```suggestion
TestConfig = collections.namedtuple("TestConfig", ["input_str", "expected", "name"])
```

or 


```suggestion
Example = collections.namedtuple("Example", ["input_str", "expected", "name"])
```

apps/librelingo_json_export/tests/test_challenges_clean_word.py
```diff
@@ -1,29 +1,27 @@
...
+        TCData("Mother-in-law", "Mother-in-law", "remove_parts_of_word_1"),
+        TCData("It's him", "It's him", "remove_parts_of_word_2"),
+        TCData(
+            "It's his mother-in-law!",
+            "It's his mother-in-law",
+            "removes_exclamation_mark",
+        ),
+        TCData("cats'", "cats'", "weird_english_posessive"),
+    ],
+    ids=lambda cur_test_case_data: cur_test_case_data.name,
```
kantord, line 24:
```suggestion
    ids=lambda test_case_data: test_case_data.name,
```

apps/librelingo_json_export/tests/test_challenges_clean_word.py
vil02: > hmm it's weird for me to create a namedtuple just for this. though I definitely see the value in having fixed labels/names for each field instead of destructuring them "blindly"
> 
> that being said it still feels weird.
> 
> have you taken this idea from somewhere? thinking

Why _just for this_? I think we both agree that seeing a direct link between the current test data set and its _name_ is extremely valuable (some of the test cases in the original code had meaningful names). The naive way of using `@pytest.mark.parametrize` with specifying the `ids` is very tedious.

Defining a `TestCaseDataType` and having a _list_ of objects of it is appears to me as a standard way of parametrising tests in `GTest` in `C++`.

I think that using a `namedtuple` here is a good compromise between using a standard `tuple` and defining a proper class (which would be an overkill, since... there is `namedtuple`).

kantord: nice, I really like it now!

PR #1999, opened 2022-01-14
fix: avoid potential bugs

vil02:
I did some basic linting of the shell scripts.

I think that it might be a good idea to:
- add `set -euo pipefail` at the beginning of every 'sh' file,
- add `shellcheck` into CI.


kantord:
> add set -euo pipefail at the beginning of every 'sh' file,
add shellcheck into CI.

I agree with you on both!

kantord (approved):

PR #2010, opened 2022-01-18
feat: use shellcheck in CI

vil02:
## Description

- I did some _basic linting_ of the shell scripts,
- I added `set -euo pipefail` to every script,
- I added [`checkShellScripts.sh`](https://github.com/vil02/LibreLingo/blob/34c1a8c49e935b737a6a5e4f8f11cff5e65ecdf4/scripts/checkShellScripts.sh), which runs `shellcheck` and verifies if every shell script contains a substring `set -euo pipefail` (well... why would one _cheat_ here?)
- I added [a workflow running](https://github.com/vil02/LibreLingo/blob/a27ac46849e84efea13bbab8b69401d0778f001c/.github/workflows/check_shell_scripts.yml) `checkShellScripts.sh`.

### DIFF
--- a/.github/workflows/check_shell_scripts.yml
+++ b/.github/workflows/check_shell_scripts.yml
@@ -1,0 +1,30 @@
+---
+name: check_shell_scripts
+
+# yamllint disable-line rule:truthy
+on:
+  workflow_dispatch:
+  push:
+  pull_request:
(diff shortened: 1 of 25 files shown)


kantord:
thank you @vil02!!!!!

Issue #2018, opened 2022-01-23
make install course use a configuration file


szabgab:
`apps/web/scripts/installExternalCourses.sh` now has a list of courses. It could use the `config/courses.json` file.

kantord:
great idea!

PR #2033, opened 2022-02-04
Implemented idea that issue #2018 suggested

cutthroat78:
In this pull request I have implemented the idea that #2018 suggested
Changes/Notes:
1. For apps/web/scripts/installExternalCourses.sh to work you need jq
2. made apps/web/scripts/installExternalCourses.sh use courses defined in config/courses.json 
Note: the script will install courses into the courses folder based on if the "dev" field is set to true (as opposed to "enable" field) because if I used the "enable" filed then only the spanish from english course would install into the courses folder
3. changed apps/web/scripts/installExternalCourse.sh to use "courses" instead of "../../courses", as "../../courses" wouldn't work for me (change this if it doesn't work you/Librelingo.app)
4. added "installCourse" that runs "./apps/web/scripts/installExternalCourse.sh" and "installAllCourses" that runs "./apps/web/scripts/installExternalCourses.sh" to package.json
5. bangla won't export as there is a mini-dictionary problem 
```
RuntimeError: Error while exporting skill "Animals": The Bangla word "আর" does not have a definition. Please add it to the mini-dictionary.
```


cutthroat78:
I can't seem to fix the checks on my own, do I need to fix anything and if so, how do I go about doing that?
Thanks in advance

szabgab:
I have not noticed you are working on this and I've also implemented it using Python. See #2036
Also I think it would be better to convert everything to Python on Node (#2035)

cutthroat78:
I wanted to use python but I wasn't sure how to go about, so I used bash instead. Thanks very much
Shall I close this pull request then?

szabgab:
Let's see what @kantord says. He is the boss.

kantord:
> I wanted to use python but I wasn't sure how to go about, so I used bash instead. Thanks very much Shall I close this pull request then?

Hey @cutthroat78 I agree, it's better to do it in Python. Let's close it in favor of @szabgab's pull request. Nice contribution anyway! Thank you! :rocket:

PR #2036, opened 2022-02-05
make install course use a configuration file closing #2018

szabgab:


apps/web/scripts/installExternalCourses.py
```diff
@@ -0,0 +1,20 @@
...
+        config = json.load(fh)
+    for course in config:
+        if course['deploy']:
+            cmd = f'''yarn installCourse "{course['url']}" {course['tdir']}'''
+            print(cmd)
+            exit_code = os.system(cmd)
+            if exit_code != 0:
+                exit("We failed to deploy {course['tdir']}. Terminating")
+
+main()
```
kantord, line 19:
```suggestion
if __name__ == '__main__':
    main()
```

kantord: see: https://stackoverflow.com/questions/419163/what-does-if-name-main-do

apps/web/scripts/installExternalCourses.py
```diff
@@ -0,0 +1,20 @@
+import json
+import sys
+import os
+
+def main():
```
kantord, line 5: I wonder, now that we are getting sophisticated enough to write Python code for this functionality, is there a way we could perhaps use some existing solution for this problem.

Looks like this is basically a package management issue so I'm wondering whether it would be a good idea to use an existing package manager for this? :thinking:

I think that the courses could be published to PyPI probably :thinking:  the only issue I see perhaps is that we would need to keep updating the lockfile in order to deploy the latest version. But if there's a non-hacky way to deal with that the it should be ok :thinking:

apps/web/package.json
```diff
@@ -12,7 +12,7 @@
     "jest": "jest src",
     "types": "sapper build && tsc && svelte-check --treshold error",
     "installCourse": "./scripts/installExternalCourse.sh",
-    "installAllExternalCourses": "./scripts/installExternalCourses.sh",
+    "installAllExternalCourses": "python ./scripts/installExternalCourses.py ../../config/courses.json",
```
kantord, line 15:
```suggestion
    "installAllExternalCourses": "poetry python ./scripts/installExternalCourses.py ../../config/courses.json",
```

kantord: btw if we are all the way going to re-writing scripts in Python (which would probably make sense, if we want to support (simple) Windows setups) then poetry scripts could be used for this perhaps, instead of yarn :thinking:

kantord: or else I'm also considering using a Makefile, or similar solution to have a "central" script runner in non-hacky ways. At least using `yarn` to run Python scripts is something some Python developers I know would take issue with :sweat_smile: What do you think?

apps/web/scripts/installExternalCourses.py
```diff
@@ -0,0 +1,20 @@
...
+    if len(sys.argv) != 2:
```
kantord, line 6: what do you think about using the[ click library](https://click.palletsprojects.com/en/8.0.x/) for this? I know this is a simple script, but click does not need boilerplate at all. It would be like this basically:

```python
import click

@click.command()
@click.argument('config_file')
def main(config_file):
    pass
```

it does the argument parsing for you and even generates the error messages and `--help`

apps/web/scripts/installExternalCourses.py
```diff
@@ -0,0 +1,20 @@
...
+        exit(f"Usage: {sys.argv[0]} PATH_TO_CONFIG_FILE")
+    filename = sys.argv[1]
```
kantord, line 8:
```suggestion
    config_filename = sys.argv[1]
```

I would argue that filename is not a great variable name because it more or less just describes the "type" of the value, not it's purpose or role

apps/web/scripts/installExternalCourses.py
```diff
@@ -0,0 +1,20 @@
...
+    with open(filename) as fh:
```
kantord, line 9: why `fh`? :thinking: 

It's not a big deal here as this is a pattern any Python developer will recognize without thinking, but I think abbreviations and 1-2 character variable names are generally not great. :thinking: for one I can't figure out what `fh` stands for, even though I understand the code

apps/web/scripts/installExternalCourses.py
```diff
@@ -0,0 +1,20 @@
...
+        config = json.load(fh)
+    for course in config:
```
kantord, line 11: I would skip an empty line here. I wonder if there's a way to configure our linter tools to enforce this :thinking:

apps/web/scripts/installExternalCourses.py
```diff
@@ -0,0 +1,20 @@
...
+        if course['deploy']:
+            cmd = f'''yarn installCourse "{course['url']}" {course['tdir']}'''
```
kantord, line 13: we should probably rewrite this fully in Python then, no?

apps/web/scripts/installExternalCourses.py
```diff
@@ -0,0 +1,20 @@
...
+            print(cmd)
+            exit_code = os.system(cmd)
+            if exit_code != 0:
+                exit("We failed to deploy {course['tdir']}. Terminating")
```
kantord, line 17:
```suggestion
                exit("We failed to install {course['tdir']}. Terminating")
```

as techically this script can either run locally, or it can run on the server during the deployment process

apps/web/scripts/installExternalCourses.py
kantord: (which is of course not to say that it needs to be done in this merge request per se)

kantord (changes requested):
Hi @szabgab thank you for this pull request.

It may seem like I commented the :hankey:  out of it but actually it's almost mergeable as it is, I just wanted to point out some ideas and thoughts. It's largely up to you what you implement from it

config/courses.json
```diff
@@ -3,62 +3,71 @@
     "url": "https://github.com/LibreLingo/LibreLingo-ES-from-EN/archive/refs/heads/main.zip",
     "sdir": "LibreLingo-ES-from-EN-main",
     "tdir": "spanish-from-english",
+    "deploy": true,
```
kantord, line 6: oh and a last question, what's the difference between `"deploy"` and `"enable"`?

kantord: also, my understanding is that there's currently no difference between a course being deployed vs installed. It's just about the context, if someone is setting up their dev environment, then it's install, but during the automatic deployment process it's deploy.

Other than the naming though, it makes a lot of sense. But maybe then we can also consider the naming of the script itself:

/scripts/installExternalCourses.py  -> /scripts/setupCoursesForDeployment.py

or maybe, support 2 different use cases? Development vs deployment.

```bash
/scripts/installExternalCourses.py production  # Installs all courses that are meant to be deployed in production
/scripts/installExternalCourses.py development  # Installs all courses that are meant to be installed in a local environment
```

what do you think? I'm not sure if it makes sense. But in that case I perhaps imagine the config file having this syntax:

```suggestion
  "enabled": {
    "development": true,
    "production": true
  }
```

kantord: btw I remember from our video call that you mentioned that this process is missing from the documentation: https://librelingo.app/docs/

that happened because I forgot to update the documentation when this install logic was introduced. Perhaps this Pull Request is the ideal time to put this feature in the documentation.

You can change this file in order to document your script: https://github.com/LibreLingo/LibreLingo/blob/main/docs/index.md

szabgab: dev: generate for the dev-tool (we might occasionally need to disable something even there)
deploy: put the json files on the production server
enable: show on the front-page (this is not used yet)

kantord:
Oh and one more thing. It would be nice to have unit tests for this. That's one of the main advantages of switching from shell scripts to bash, so let's take advantage of it!

apps/web/scripts/installExternalCourses.py
szabgab: Oh I know what it does, I am teaching this stuff to innocent students :)

config/courses.json
kantord: so I suggest something like this:

```json
  "enabled": {
    "development": true,
    "production": true
  }
```

just for clarity. it's better if it's self-explanatory, rather than having to document it and make people read the documentation. (Not that it cannot be mentioned in docs anyway)

apps/web/scripts/installExternalCourses.py
szabgab: I don't think we need to assume that the course authors know PyPI or even Python.

apps/web/package.json
szabgab: Makefile would require some kind of a make command on your computer :) Not a very Windowsy thing.

szabgab: The current script only uses standard-Python modules so there is not a lot of need for poetry. For now.

apps/web/scripts/installExternalCourses.py
szabgab: Maybe another time.

config/courses.json
kantord: btw I think that the same config file should be used for implementing the landing page and in general course selectors. so basically we will need to have some config also to configure whether a course should be shown in course selectors (the rest will be only shown in the dev tools)

apps/web/scripts/installExternalCourses.py
szabgab: I am fighting for moving from f to fh at a client, it is already quite searchable and quite acceptable, but if you'd like some other name, I open for it.

apps/web/scripts/installExternalCourses.py
szabgab: Well, yes, but I prefer to make changes to be smaller rather bigger.

config/courses.json
szabgab: That's what the "enable" was intended for.

apps/web/scripts/installExternalCourses.py
kantord: that's true, but the packaging can be done automatically, similarly how linting can be done automatically. Basically, the CI tool can automatically publish a new version on the `main` branch if no issues were detected

apps/web/package.json
kantord: basically poetry scripts are just python commands that are added to path. for sure that's a lot less flexible than what yarn does

kantord: > Makefile would require some kind of a make command on your computer :) Not a very Windowsy thing.

that's a good point

apps/web/scripts/installExternalCourses.py
kantord: I would just ditch abbreviations.

```python
with open(filename) as config_file:
```

Thought this is not the most critical example of abbreviations being bad

apps/web/scripts/installExternalCourses.py
kantord: yes, I agree with that

config/courses.json
kantord: perhaps `"public"` would be a better name for that? :thinking:

kantord:
Hey @szabgab  I noticed that there's also a pylint failure" https://kantord.semaphoreci.com/jobs/d735f029-46fc-46a4-aec5-c76a32e6473f

PR #2040, opened 2022-02-10
chore(pylint): remove trailing unused-argument commented pylint rule

qbecb1zen:
removed trailing pylint commented rule


kantord:
thanks!

PR #2056, opened 2022-02-20
Qbecb1zen/improv gitignore

qbecb1zen:
- reorganize `.gitignore`
- enable pylint rule `line-too-long`


qbecb1zen:
(i am giving the failing tests a look)

apps/librelingo_types/librelingo_types/data_types.py
```diff
@@ -78,6 +79,7 @@ class HunspellSettings(
     pass
 
 
+# pylint: disable=line-too-long
```
qbecb1zen, line 82: i disable them here, because i didn't found an "elegant" solution for the docstring tests, if you have any i am eager to discover it!

qbecb1zen:
@kantord do you have an idea on why cypress and install step for black failed ? (it seem to me like a random failure as locally it's working and i didn't touched to any installation step for this PR)

kantord:
> @kantord do you have an idea on why cypress and install step for black failed ? (it seem to me like a random failure as locally it's working and i didn't touched to any installation step for this PR)

it could be random failures indeed. with cypress it's somewhat understandable, as it's testing a very complex thing but with the poetry dependency install, that's weird. it should not randomly fail under any circumstances :thinking:

kantord:
thank you @qbecb1zen !

PR #2058, opened 2022-02-20
add makefile with a few commands

qbecb1zen:
This PR is here to open a few discussions concerning some points i was thinking of (yes i just had some extra time haha)

But can we get rid of those `git ls-files '*.py'` from the CI and use some `$(find...)` command, it will not make a big change, but for the next contributors i feel like they will be using this command to simulate the CI whenever it fails, which can lead to some errors on their side (just dev UX), as this command will fail whenever a file is added or removed and is not commited.

A second point, is to have a makefile with useful commands, also a help command that will generate the documentation of those commands, with the pushed makefile we have something similar to:

Screenshot 2022-02-20 at 11 40 39
 
that is pretty useful for users not used for the codebase (and it's useful just to find quickly the needed commands)

I also added cache deletion commands if anyone need them :) ! 

Eager to have your returns on each and if anyone is reluctant :)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,4 +21,4 @@ backend-ci: ## Run python corresponding ci
 	make format && make lint && make type-check
 
 help:
-	grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-25s\033[0m %s\n", $$1, $$2}'
+	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-25s\033[0m %s\n", $$1, $$2}'
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,24 @@
+PYTHON_FILEPATHS = `(find . -iname "*.py" -not -path "./.venv/*")`
+lint: ## Lint codebase
+	poetry run pylint --rcfile=pylintrc $(PYTHON_FILEPATHS)
+
+format: ## Run black formatter
+	poetry run black --check $(PYTHON_FILEPATHS)
+
+format-fix: ## Run black formatter with automated fix
+	poetry run black $(PYTHON_FILEPATHS)
+
+type-check: ## Run black formatter with automated fix
+	poetry run mypy $(PYTHON_FILEPATHS)
+
+pycache-delete: ## Delete the __pycache__ folders
+	find . -type d -name __pycache__ -exec rm -r {} \+
+
+clean: ## Delete cache generated
+	rm -fr .mypy_cache/ .pytest_cache/ && make pycache-delete
+
+backend-ci: ## Run python corresponding ci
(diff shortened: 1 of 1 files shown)


kantord:
I'm not so much against Makefile, but it seems like we already have 2 systems that are able to run scripts (`poetry` and `yarn`). Is there a way maybe we could use those to achieve the same features?

Or else, is there some alternative idea that we can consider?

Makefile
```diff
@@ -0,0 +1,24 @@
...
+	find . -type d -name __pycache__ -exec rm -r {} \+
+
+clean: ## Delete cache generated
+	rm -fr .mypy_cache/ .pytest_cache/ && make pycache-delete
+
+backend-ci: ## Run python corresponding ci
+	make format && make lint && make type-check
+
+help:
+	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-25s\033[0m %s\n", $$1, $$2}'
```
kantord, line 24: Interesting solution!

qbecb1zen: yes it's a solution i made once for a personal project, since whenever i show it to people, they love it haha

qbecb1zen:
> I'm not so much against Makefile, but it seems like we already have 2 systems that are able to run scripts (`poetry` and `yarn`). Is there a way maybe we could use those to achieve the same features?
> 
> Or else, is there some alternative idea that we can consider?

Oh i forgot about those, we can have those commands in pyproject.toml or package.json, i just find it awkward to have inside package.json commadns with `poetry run black $(PYTHON_FILEPATHS)` (i did just python commands as i only worked in the python part for the moment, but i am more seing the makefiles as the files that will do the link between yarn/poetry (to mix commands)

qbecb1zen:
> > > I'm not so much against Makefile, but it seems like we already have 2 systems that are able to run scripts (`poetry` and `yarn`). Is there a way maybe we could use those to achieve the same features?
> > > Or else, is there some alternative idea that we can consider?
> > 
> > 
> > Oh i forgot about those, we can have those commands in pyproject.toml or package.json, i just find it awkward to have inside package.json commadns with `poetry run black $(PYTHON_FILEPATHS)` (i did just python commands as i only worked in the python part for the moment, but i am more seing the makefiles as the files that will do the link between yarn/poetry (to mix commands)
> 
> yes, it's indeed weird to run python scripts from package.json scripts 😅

That's why i was suggesting those makefiles, so they can make the link between both poetry and yarn commands!

qbecb1zen:
Ah i always forgot about Windows developers 🤓🙄😏!

Also if this is not better for the dev experience we can just close the PR, i was more thinking as i joined the project recently that it may be better for dev experience to have a place where useful commands are so everything is centralized and easier for new people joining LibreLingo contributors haha!

kantord:
So how about we say, Windows development is supported through WSL as well as Docker. OSX and Linux supported with or without Docker

kantord:
I added an issue for documenting it: https://github.com/LibreLingo/LibreLingo/issues/2061

kantord:
Thanks @qbecb1zen !

PR #2062, opened 2022-03-03
Changed Docker image from nodejs to Arch Linux

cutthroat78:
Changes:
- Made the image smaller
- Made the dockerfile much neater
- Simplified installation of packages in dockerfile
- Updated readme
- Allow for the importing of multiple courses and courses can have whatever name rather than just having "your-course"
- Moved image name on hub.docker.com from cutthroat78/librelingo to librelingo/librelingo to make an easier image name to remember

### DIFF
--- a/apps/docker/Dockerfile
+++ b/apps/docker/Dockerfile
@@ -3,9 +3,11 @@ FROM archlinux:base-20220227.0.49015
 # Setup packages
 RUN pacman -Syu --noconfirm git nodejs-lts-fermium gcc yarn python3 python-pip
 
-# Setup Poetry and LibreLingo Python Packages
-RUN curl -sSL https://install.python-poetry.org | python3 - && \
-  cp /root/.local/bin/poetry /bin
+# Setup Poetry
+RUN curl -sSL https://install.python-poetry.org | python3 -
+
+# Update path to include poetry
+ENV PATH="/root/.local/bin/:${PATH}"
   
 # Git clone LibreLingo repo
 RUN git clone https://github.com/LibreLingo/LibreLingo.git
--- a/apps/docker/Dockerfile
+++ b/apps/docker/Dockerfile
@@ -1,38 +1,27 @@
-# Librelingo is not compatible with Node 16, using Node 14.18.3 image instead
-FROM node:14.18.3 
+FROM archlinux:base-20220227.0.49015
 
-# Setup Librelingo
-RUN git clone https://github.com/LibreLingo/LibreLingo.git
-WORKDIR /LibreLingo
-RUN yarn set version classic && \
-  yarn
-
-# Setup python, poetry and python modules to be able to convert courses from YAML to JSON
-RUN wget https://www.python.org/ftp/python/3.9.0/Python-3.9.0.tar.xz && \
-  tar xf Python-3.9.0.tar.xz && \
-  cd ./Python-3.9.0 && \
-  ./configure && \
-  make && \
-  make install && \
-  update-alternatives --install /usr/bin/python python /usr/local/bin/python3.9 10 && \
-  ln -s /usr/share/pyshared/lsb_release.py /usr/local/lib/python3.9/site-packages/lsb_release.py && \
-  pip3 install --upgrade pip && \
-  pip install pipx && \
-  pipx install poetry && \
-  export PATH=$PATH:/root/.local/bin && \
-  poetry install && \
-  cd /LibreLingo/apps/librelingo_yaml_loader && \
-  poetry install && \
-  cd /LibreLingo/apps/librelingo_json_export && \
-  poetry install && \
+# Setup packages and Librelingo
+RUN pacman -Syu --noconfirm git nodejs-lts-fermium gcc yarn python3 python-pip && \
+  curl -sSL https://install.python-poetry.org | python3 - && \
   pip install librelingo-yaml-loader && \
-  pip install librelingo-json-export
+  pip install librelingo-json-export && \
+  git clone https://github.com/LibreLingo/LibreLingo.git && \
+  cd /LibreLingo && \
+  yarn set version classic && \
+  yarn && \
+  /root/.local/bin/poetry install && \
+  cd /LibreLingo/apps/librelingo_yaml_loader && \
+  /root/.local/bin/poetry install && \
+  cd /LibreLingo/apps/librelingo_json_export && \
+  /root/.local/bin/poetry install
 
-# Entrypoint script that will convert the course at /LibreLingo/courses/your-course from YAML to JSON and make it available in Librelingo Web App copy and chmod
-COPY start.sh /LibreLingo/start.sh
+WORKDIR /LibreLingo
+
+# Entrypoint script that will convert all the courses in /LibreLingo/courses and make it available in Librelingo
+COPY docker-entrypoint.sh /LibreLingo/docker-entrypoint.sh
 RUN chmod +x /LibreLingo/start.sh
 
 EXPOSE 3000
 
-# Run entrypoint script
-ENTRYPOINT ["./start.sh"]
+# Run entrypoint script as the entrypoint
+ENTRYPOINT ["/LibreLingo/docker-entrypoint.sh"]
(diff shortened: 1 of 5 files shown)


apps/docker/Dockerfile
```diff
@@ -1,38 +1,37 @@
-# Librelingo is not compatible with Node 16, using Node 14.18.3 image instead
-FROM node:14.18.3 
+FROM archlinux:base-20220227.0.49015
```
kantord, line 1: this way, we'll have to keep updating this a lot, no? is there a good mechanism to keep this up to date? at least with the node image, it's mainly about updating major node versions :thinking: (although I can see, we had a very specific tag there as well)

apps/docker/Dockerfile
```diff
@@ -1,38 +1,37 @@
...
 
-# Setup Librelingo
+# Setup packages
+RUN pacman -Syu --noconfirm git nodejs-lts-fermium gcc yarn python3 python-pip
+
+# Setup Poetry and LibreLingo Python Packages
+RUN curl -sSL https://install.python-poetry.org | python3 - && \
+  cp /root/.local/bin/poetry /bin && \
+  pip install librelingo-yaml-loader && \
```
kantord, line 9: we should not need this, and the next line either. we should be able to install these packages from the local repo as opposed to PyPI in the Docker image

apps/docker/Dockerfile
```diff
@@ -1,38 +1,37 @@
...
+  pip install librelingo-json-export
+  
+# Git clone LibreLingo repo
 RUN git clone https://github.com/LibreLingo/LibreLingo.git
```
kantord, line 13: actually, I realize now that we should not git clone the repo in the Dockerfile, right? I mean, we should simply add the relevant files from the folder the user already has, so that local changes can also be tested using Docker?

This PR can be accepted without that fix though, because this "mistake" already existed before

apps/docker/Dockerfile
```diff
@@ -1,38 +1,37 @@
...
-WORKDIR /LibreLingo
-RUN yarn set version classic && \
+  
+# Setup LibreLingo 
+RUN cd /LibreLingo && \
+  yarn set version classic && \
   yarn
```
kantord, line 18: this should probably be

```bash
yarn install --frozen-lockfile
```

apps/docker/Dockerfile
```diff
@@ -1,38 +1,37 @@
...
-  ln -s /usr/share/pyshared/lsb_release.py /usr/local/lib/python3.9/site-packages/lsb_release.py && \
-  pip3 install --upgrade pip && \
-  pip install pipx && \
-  pipx install poetry && \
-  export PATH=$PATH:/root/.local/bin && \
-  poetry install && \
+ 
+# Setup tools to import courses
+RUN cd /LibreLingo && \
+  /root/.local/bin/poetry install && \
```
kantord, line 22: this is weird, why doesn't this work? 

```suggestion
  poetry install && \
```

apps/docker/Dockerfile
```diff
@@ -1,38 +1,37 @@
...
   cd /LibreLingo/apps/librelingo_yaml_loader && \
-  poetry install && \
+  /root/.local/bin/poetry install && \
   cd /LibreLingo/apps/librelingo_json_export && \
-  poetry install && \
-  pip install librelingo-yaml-loader && \
-  pip install librelingo-json-export
+  /root/.local/bin/poetry install
+
+WORKDIR /LibreLingo
```
kantord, line 28: this could have been done before, no?

kantord (changes requested):

apps/docker/Dockerfile
cutthroat78: A specific tag is used so that if anyone runs the dockerfile they will get the same image as you would get on docker hub. Using no tag could cause problems for some people running the dockerfile as the new arch image used might have changes that make other parts of the dockerfile not work

cutthroat78: Though, I might change the docker to use a distro with a package manager where I can specify specific version of packages, so that there won't be any problems when running the dockerfile when there are newer packages than the ones used in the image on docker hub

cutthroat78: My problem with the node image was the finicky way of having to setup pip and python 3

apps/docker/Dockerfile
cutthroat78: I will change it, to use the local ones when I have time

apps/docker/Dockerfile
cutthroat78: My original idea for the docker image was to test courses in Librelingo by just running one command that would setup the docker container and volume mount in, but I do agree that the docker image should be used to help work on development on librelingo. Maybe, make two docker images. One with the git clone for course testing in Librelingo and another for development

apps/docker/Dockerfile
cutthroat78: I will change that when I have the chance. I have not much experience with yarn, so I didn't know. Thanks!

apps/docker/Dockerfile
cutthroat78: I couldn't figure a way to add /root/.local/bin/ to the $PATH in a dockerfile but this line was also the help this issue ```cp /root/.local/bin/poetry /bin``` as well. The line I specified there was also for the docker-entrypoint.sh file

apps/docker/Dockerfile
cutthroat78: I am not sure which part you are talking about line 26, 29 or line 28

apps/docker/Dockerfile
cutthroat78: I found a way to update the $PATH by using "ENV PATH="/root/.local/bin/:${PATH}""

apps/docker/Dockerfile
cutthroat78: I might use an alpine image as alpine images are small and it's package manager allows you to specify package version but alpine images/containers can be a bit hard to work with though

apps/docker/Dockerfile
cutthroat78: I could perhaps setup a label that allows you to set if do or don't want the repo git cloned when the docker-entrypoint.sh runs

apps/docker/Dockerfile
cutthroat78: I am not sure what you are pointing but I have moved the WORKDIR to after the git clone and got rid of the unncessary cd commands

apps/docker/Dockerfile
cutthroat78: Changed!

cutthroat78:
I have made some changes but they have not been tested yet

cutthroat78:
I have made some changes based of @kantord suggestions and my own suggestion and have tested them. Any other suggestions or ideas?

apps/docker/Dockerfile
kantord: hmmm, right, it makes sense to not use the `node` image, that's for sure.  yeah the specific tag is needed for sure, I mean we need to have a specific version of node and python

btw what about having 2 different dockerfiles for python stuff and node stuff? would that not make things much simpler?

kantord: ok, using alpine seems more standard I think, so I like that idea

apps/docker/Dockerfile
kantord: amazing!

apps/docker/Dockerfile
kantord: amazing!

kantord:
I think it's good now, at the very least good enough to merge right now, thanks @cutthroat78

kantord:
@allcontributors please add @cutthroat78 for code

kantord:
@allcontributors please add @cutthroat78 for code

cutthroat78:
I will change the image over to alpine and set an environment variable to enable/disable python commands when I have the chance. I am just a bit busy at the moment.

Thanks!

kantord:
@allcontributors please add @cutthroat78 for code

PR #2067, opened 2022-03-10
feat: allow the creation of two-way dictionaries 

szabgab:
Instead of adding a pair of words in both direction of the mini-dictionary, this will allow adding the pair only once and it will add the entry in both directions.

### DIFF
--- a/apps/librelingo_yaml_loader/librelingo_yaml_loader/yaml_loader.py
+++ b/apps/librelingo_yaml_loader/librelingo_yaml_loader/yaml_loader.py
@@ -220,6 +220,23 @@ def _convert_phrases(raw_phrases) -> List[Phrase]:
     return list(map(_convert_phrase, raw_phrases))
 
 
+def _convert_two_way_dictionary(data, course: Course):
+    """
+    Handles loading the twoway-dictionary form the YAML format
+    """
+    dictionary = []
+    # return dictionary
+    if "Twoway-dictionary" not in data:
+        return dictionary
+    raw_two_way_dictionary = data["Twoway-dictionary"]
+    for item in raw_two_way_dictionary:
+        word = list(item.keys())[0]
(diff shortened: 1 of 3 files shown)


szabgab:
I don't know what do others have in their mini-dictionaries. I think I'd convert all of my mini dictionaries into these two-way dictionaries, but I would be afraid just changing how the mini-dictionaries work today. However if people like this two-way dictionary then at one point we might deprecate the mini-dictionary.

I'd say let's add this and then see what happens and what other editors think.

apps/librelingo_yaml_loader/librelingo_yaml_loader/yaml_loader.py
```diff
@@ -220,6 +220,23 @@ def _convert_phrases(raw_phrases) -> List[Phrase]:
...
 
+def _convert_two_way_dictionary(data) -> List[tuple]:
+    """
+    Handles loading the twoway-dictionary form the YAML format
+    """
+    dictionary: List[tuple] = []
+    # return dictionary
+    if "Twoway-dictionary" not in data:
+        return dictionary
+    raw_two_way_dictionary = data["Twoway-dictionary"]
```
kantord, line 231:
```suggestion
    raw_two_way_dictionary = data["Two-way dictionary"]
```

then the only remaining thing I'd suggest is using this spelling

kantord (changes requested):

kantord:
@szabgab one thing that would be nice is to have some documentation for this. There's already documentation for the mini dictionary here: https://github.com/LibreLingo/LibreLingo/blob/main/docs/courses/skill.md#data-breakdown

szabgab:
@kantord , sure I am going to add that too. In the meantime I think my CI fails as it needs a new release of the yamlreader package. Could you release it please?

Issue #2076, opened 2022-03-18
Test the "test course"


szabgab:
Create a test that will load the `courses/test` and compare the results with `apps/web/src/courses/test/`

PR #2077, opened 2022-03-18
Test with test course

szabgab:
Implementing  #2076

### DIFF
--- a/apps/tools/tests/test_test_course.py
+++ b/apps/tools/tests/test_test_course.py
@@ -1,0 +1,57 @@
+import collections
+import json
+import os
+
+from librelingo_json_export.export import export_course
+from librelingo_yaml_loader.yaml_loader import load_course
+
+Settings = collections.namedtuple(
+    "Settings",
+    [
+        "dry_run",
+    ],
+)
+
+
+def test_exported_test_course_matches_snapshot(tmpdir):
+    root = os.path.abspath(__file__)
+    root = os.path.dirname(root)
+    root = os.path.dirname(root)
+    root = os.path.dirname(root)
+    root = os.path.dirname(root)
+
+    path_to_course = os.path.join(root, "courses", "test")
+    course = load_course(path_to_course)
+
+    settings = Settings(
+        dry_run=False,
+    )
+    output_path = str(tmpdir)
+    expected_path = os.path.join(root, "apps", "web", "src", "courses", "test")
+
+    export_course(output_path, course, settings)
+    assert_folders_are_identical(expected_path, output_path)
+
+
+def assert_folders_are_identical(expected_path, actual_path):
+    for dirname, _, files in os.walk(actual_path):
+        for filename in files:
+            relpath = os.path.join(dirname[len(actual_path) + 1 :], filename)
+
+            if filename.endswith(".json"):
+                with open(os.path.join(actual_path, relpath)) as fh:
+                    actual = json.load(fh)
+                with open(os.path.join(expected_path, relpath)) as fh:
+                    expected = json.load(fh)
+                assert actual == expected, relpath
+                continue
+
+            if filename.endswith(".md"):
+                with open(os.path.join(actual_path, relpath)) as fh:
+                    actual = fh.read()
+                with open(os.path.join(expected_path, relpath)) as fh:
+                    expected = fh.read()
+                assert actual == expected, relpath
+                continue
+
+            assert False, f"Unhandled file {relpath}"


apps/tools/tests/test_test_course.py
```diff
@@ -0,0 +1,55 @@
...
+
+    path_to_course = os.path.join(root, "courses", "test")
+    course = load_course(path_to_course)
+
+    settings = Settings(
+        dry_run=False,
+    )
+    output_path = str(tmpdir)
+    expected_path = os.path.join(root, "apps", "web", "src", "courses", "test")
+    print(output_path)
```
kantord, line 32:
```suggestion
```

apps/tools/tests/test_test_course.py
```diff
@@ -0,0 +1,55 @@
...
+
+Settings = collections.namedtuple(
+    "Settings",
+    [
+        "dry_run",
+    ],
+)
+
+
+def test_course(tmpdir):
```
kantord, line 16:
```suggestion
def test_exported_test_course_matches_snapshot(tmpdir):
```

apps/tools/tests/test_test_course.py
```diff
@@ -0,0 +1,55 @@
...
+    root = os.path.abspath(__file__)
+    root = os.path.dirname(root)
+    root = os.path.dirname(root)
+    root = os.path.dirname(root)
+    root = os.path.dirname(root)
+    print(root)
```
kantord, line 22:
```suggestion
```

apps/tools/tests/test_test_course.py
```diff
@@ -0,0 +1,55 @@
...
+    ],
+)
+
+
+def test_course(tmpdir):
+    root = os.path.abspath(__file__)
+    root = os.path.dirname(root)
+    root = os.path.dirname(root)
+    root = os.path.dirname(root)
+    root = os.path.dirname(root)
```
kantord, line 21: is there a different way to get this root path, I'm thinking? and if not, perhaps there's a way to put it into an automatically applied pytest fixture?

apps/tools/tests/test_test_course.py
```diff
@@ -0,0 +1,55 @@
...
+    print(root)
+
+    path_to_course = os.path.join(root, "courses", "test")
```
kantord, line 24: I am thinking, this path could be moved to a pytest fixture, for example a fixture called `test_course_snapshot`

apps/tools/tests/test_test_course.py
```diff
@@ -0,0 +1,55 @@
...
+
+            if filename.endswith(".md"):
+                with open(os.path.join(output_path, relpath)) as fh:
+                    actual = fh.read()
+                with open(os.path.join(expected_path, relpath)) as fh:
+                    expected = fh.read()
+                assert actual == expected, relpath
+                continue
+
+            assert False, f"Unhandled file {relpath}"
```
kantord, line 55: it could even be a simple exception, does not need to be an assertion. but it's good like this too

apps/tools/tests/test_test_course.py
```diff
@@ -0,0 +1,55 @@
...
+
+    settings = Settings(
+        dry_run=False,
+    )
+    output_path = str(tmpdir)
+    expected_path = os.path.join(root, "apps", "web", "src", "courses", "test")
+    print(output_path)
+
+    export_course(output_path, course, settings)
+    for dirname, _, files in os.walk(output_path):
```
kantord, line 35: this logic could be extracted to another function

```python
def compare_folders():
    # ... etc ...
```

kantord: or 

```python
def assert_folders_are_identical():
    # ... etc ...
```

apps/tools/tests/test_test_course.py
szabgab: It disturbs me too that dirname does not accept  a second parameter of how many levels.
I think it could be moved to a fixture only if we knew that every test file that might need this will be at the same depths in our tree.

apps/tools/tests/test_test_course.py
szabgab: but why, is in needed in other places as well?

apps/tools/tests/test_test_course.py
szabgab: Well, as I understand an assert is just an easy way to generate an exception on a condition.

kantord: sure, I just wasn't sure if it will create the proper exception class in this case or not. I was working with the assumption that this is basically signalling an issue with the test itself, not with the unit under test. But I thought I bit more and I realized that it can also be because the functionality of that function was changed and generates new files.

so it's perfect like this :100:

apps/tools/tests/test_test_course.py
kantord: that's a good point, I didn't think about that. then probably this is the best, I mean, the next best thing I could think of would be a reduce function, but I don't like reduce functions so it will be fine like this

PR #2121, opened 2022-03-31
Update image_attributions.csv

mellishim:
Adding new words. Couldn't find good images for "you all," so the same image is being used for all three until better ones are available.


kantord:
@mellishim  a tip: you don't need to find an image for "you all". Instead of teaching that as a word, you can just include it in a phrase. The users will be able to understand what it means because of the mini-dictionary.

PR #2135, opened 2022-04-08
Adds "go back" button to course introduction page

### LINKED ISSUE #2079: Add "go back" button to course introduction pages
Modify this page: https://github.com/LibreLingo/LibreLingo/blob/eba232b43ab98606a71d9253024ab50ea8a26e5c/apps/web/src/routes/course/%5BcourseName%5D/skill/%5BskillName%5D/introduction.svelte 

so that there's a way to go back to the main page of the same course. There's already a button that starts practicing the skill, so this button will be added in addition to that.

It should be very simple. Just add a button with the link, just like the practice button now. The URL is pretty simple, here's an example of a URL for a course page: https://librelingo.app/course/spanish-from-english

For comparison, here's an example of a URL of an introduction page: https://librelingo.app/course/spanish-from-english/skill/animals/introduction

rodrigo-pova:
resolves #2079

### DIFF
--- a/apps/web/src/routes/course/[courseName]/skill/[skillName]/introduction.svelte
+++ b/apps/web/src/routes/course/[courseName]/skill/[skillName]/introduction.svelte
@@ -64,14 +64,14 @@
   <MarkDownPage {readmeHTML} {title} description={$_("about.meta.description")}>
     <div>
       <Button
-        style="primary"
-        href={`course/${courseName}/skill/${practiceHref}`}
-        >Practice {title}</Button
+        style="secondary"
+        href={`course/${courseName}`}
+        >Go back to the course</Button
       >
       <Button
         style="primary"
-        href={`course/${courseName}`}
-        >Go back to the course</Button
+        href={`course/${courseName}/skill/${practiceHref}`}
+        >Practice {title}</Button
       >
     </div>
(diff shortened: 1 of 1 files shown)


kantord:
nice! :rocket: 

thank you @rodrigo-pova

apps/web/src/routes/course/[courseName]/skill/[skillName]/introduction.svelte
```diff
@@ -68,6 +68,11 @@
         href={`course/${courseName}/skill/${practiceHref}`}
         >Practice {title}</Button
       >
+      <Button
+        style="primary"
+        href={`course/${courseName}`}
+        >Go back {title}</Button
```
Comment (kantord) [inline apps/web/src/routes/course/[courseName]/skill/[skillName]/introduction.svelte:74]:
```suggestion
        >Go back to the course</Button
```

I would do it this way. And, I would also change the order of the buttons. What do you think? 

And the go back button should not be primary

Comment (rodrigo-pova) [reply apps/web/src/routes/course/[courseName]/skill/[skillName]/introduction.svelte:74]: Agreed! I'm pushing these changes in a bit

kantord:
Thank you @rodrigo-pova!

kantord:
@allcontributors please add @rodrigo-pova for code

PR #2153, opened 2022-04-14
Word pairing challenge issue

ebethwu:


ebethwu:
Some info on the failing checks: I created a pairing.feature file to give some possible scenarios when the front end is created (I know some of the syntax is incorrect, hence the failing Cypress and Percy tests). We can choose not to include the file if it causes errors. I was failing test_integration.py from the when I downloaded the initial code.

kantord:
> Some info on the failing checks: I created a pairing.feature file to give some possible scenarios when the front end is created (I know some of the syntax is incorrect, hence the failing Cypress and Percy tests). We can choose not to include the file if it causes errors. I was failing test_integration.py from the when I downloaded the initial code.

since the idea is to implement this feature soon, then (even though I'm generally against committing dead code) it might be a good idea to simple comment it out. Or perhaps renaming it from `xyz.feature` to `xyz.feature.disabled`

apps/librelingo_json_export/tests/test_challenges_pairing_challenge.py
```diff
@@ -0,0 +1,54 @@
...
+                "priority": 0,
+                "group": "b4be982a043f",
+            },
+            "data_set_2",
+        ),
+    ],
+    ids=lambda example: example.name,
+)
+def test_get_pairing_challenge(example):
+    print(example.word)
```
kantord, line 52:
```suggestion
```

it should not be necessary to print this here, instead we should rely on pytest in order to provide output that's easy to debug. I think that should already be the case, right? I mean if you run pytest, it should tell you what values `get_pairing_challenge` was called with, in case the test fails

apps/librelingo_json_export/librelingo_json_export/challenge_types.py
```diff
@@ -92,6 +92,26 @@ def get_options_challenge(phrase, _):
         }
     ]
 
+def get_pairing_challenge(words, _):
+    if not isinstance(words, list):
+        words = [words]
+    return [
```
kantord, line 98:
```suggestion

    return [
```

just for code style

apps/librelingo_json_export/tests/test_challenges_pairing_challenge.py
ebethwu: Oh yes, my mistake. It was just for debugging purposes and it looks like I forgot to remove it.

apps/librelingo_fakes/librelingo_fakes/fakes.py
```diff
@@ -113,6 +113,24 @@
     in_target_language=["apfel"], in_source_language=["apple"], pictures=["1", "2", "3"]
 )
 
+pairs1 = [Word(
```
kantord, line 116: I think this is the only "major" issue see in this PR. :thinking:  why are the variables `pairs1` and `pairs2` even needed? I mean, they are just lists of words. But skills are already lists of words. In the other challenge types fake skills are used for this purpose I think. Which is how it should be done here as well.

apps/librelingo_json_export/librelingo_json_export/challenge_types.py
```diff
@@ -92,6 +92,27 @@ def get_options_challenge(phrase, _):
...
+        {
+            "type": "pairing",
+            "pairs" : [
+                {
+                    "formInTargetLanguage": remove_control_characters_for_display(
+                        word.in_target_language[0]),
+                    "meaningInSourceLanguage": remove_control_characters_for_display(
+                        word.in_source_language[0])
+                }
+            for word in words],
```
kantord, line 109: one more thing that's missing here is filtering out invalid pairs (such as when one thing has a translation that is also the translation of something else, and similar cases).

This PR can be merged even without that, because it won't be included in the frontend yet anyway. But it would be nice if you implemented this in either this PR or a second PR :pray:

kantord (changes requested):
:100: :rocket: very good @ebethwu !!!

this is a nearly perfect PR. Just some small things that I noted in the comments

apps/librelingo_fakes/librelingo_fakes/fakes.py
ebethwu: I am having a bit of trouble trying to use fake skills. In challenges.py, I can no longer call get_pairing_challenge under _get_phrase_challenges or _get_word_challenges as I get this error "ERROR test_challenges_get_phrase_challenges.py - AttributeError: 'Phrase' object has no attribute 'words'". I am unable to find a function that will pass in a skill. Just wondering if you have any tips?

kantord: You should probably use `skillWith3Words` from fakes.py in order to achieve what you want. I think everything should work then :thinking:

ebethwu: Yes, that is what I am currently trying to use, however I am having a bit of trouble on how to call the pairing function in challenges.py now. 

Screenshot from 2022-04-17 12-29-55

apps/librelingo_json_export/librelingo_json_export/challenge_types.py
```diff
@@ -92,6 +92,27 @@ def get_options_challenge(phrase, _):
         }
     ]
 
+def get_pairing_challenge(words, _):
+    if not isinstance(words, list):
```
kantord, line 96: I realized that this part is odd. This should not be needed whatsoever :thinking: Instead, the tests should always operate on a set of words to begin with, therefore it should always be called with a list of words

kantord: This could be partially the reason for the errors that we are seeing, because this magic here might be creating some confusion that makes it harder to debug what's happening there

apps/librelingo_json_export/librelingo_json_export/challenges.py
```diff
@@ -57,7 +58,12 @@ def _get_phrase_challenges(phrase, course: Course):
 def _get_word_challenges(word, course: Course):
     "Generate challenges based on a Word"
     return _challenge_mapper(
-        [get_cards_challenge, get_short_input_challenge, get_listening_challenge]
+        [
+            get_cards_challenge,
+            get_short_input_challenge,
+            get_listening_challenge,
+            get_pairing_challenge,
```
kantord, line 65: I realized that this is wrong for a profound reason. 

This is my fault because I didn't remember correctly how these things actually work. Basically, all current challenges work based on individual words or phrases, but this new one doesn't.

I'm working on a change in order to make this easier to work with.

ebethwu: I was thinking that was an issue, which I think I tried to create pairs and the odd list thing to work around the fact that only a single word/phrase could be passed through. Thank you for working on a change to make it a little easier to work on!

kantord:
@ebethwu just so that I know, are you working for some kind of deadline in order to merge this PR? If I remember correctly you are not required to merge it by a deadline, just have it submitted, right (i. e. you have already completed your assignment?)

apps/librelingo_json_export/librelingo_json_export/challenges.py
kantord: I made some changes and now it seems to work well. There is still some tests that fail but that seems like is because they simply need to be updated to include this new challenge type.

I pushed my changes into this branch on my repo: https://github.com/LibreLingo/LibreLingo/commits/word-pairing-challenge

i. e. the same branch name as you have in your fork. You can merge my changes into your fork in order to get my changes on your local branch. (it should be sth like `git merge upstream/word-pairing-challenge`)

in order to do this probably you need to have `upstream` (or any other name that you choose) which is the git link for my version of the repo. If you don't have the `upstream` remote server set up yet,  the way you can do that is this command `git remote add upstream git@github.com:LibreLingo/LibreLingo.git`

ebethwu:
My assignment report is already completed, if my code is merged by Friday I receive a small bit of extra credit, however I do not want to rush the process.

ebethwu:
All the test cases now run locally. I fixed the test_challenges_get_challenges_data test case error. I added "challenge5" and "challenge6" to fakes.py and a "mock_get_skill_challenges" and corrected "test_returns_correct_challenges" function.

kantord:
> All the test cases now run locally. I fixed the test_challenges_get_challenges_data test case error. I added "challenge5" and "challenge6" to fakes.py and a "mock_get_skill_challenges" and corrected "test_returns_correct_challenges" function.

nice!

kantord:
> > All the test cases now run locally. I fixed the test_challenges_get_challenges_data test case error. I added "challenge5" and "challenge6" to fakes.py and a "mock_get_skill_challenges" and corrected "test_returns_correct_challenges" function.
> 
> nice!

looks like there are still some stuff failing. Do you need help fixing them? Perhaps it's easier if we have a call and share screens to work together

ebethwu:
> > > All the test cases now run locally. I fixed the test_challenges_get_challenges_data test case error. I added "challenge5" and "challenge6" to fakes.py and a "mock_get_skill_challenges" and corrected "test_returns_correct_challenges" function.
> > 
> > 
> > nice!
> 
> looks like there are still some stuff failing. Do you need help fixing them? Perhaps it's easier if we have a call and share screens to work together

What is still failing?

kantord:
> > > > All the test cases now run locally. I fixed the test_challenges_get_challenges_data test case error. I added "challenge5" and "challenge6" to fakes.py and a "mock_get_skill_challenges" and corrected "test_returns_correct_challenges" function.
> > > 
> > > 
> > > nice!
> > 
> > 
> > looks like there are still some stuff failing. Do you need help fixing them? Perhaps it's easier if we have a call and share screens to work together
> 
> What is still failing?

check out this link to see the report from the pipeline: https://kantord.semaphoreci.com/workflows/6a7d77b9-326b-4369-863c-23ec1e90b338?pipeline_id=4b77fa46-d4ea-4633-ac13-9f1a6b53de19

PR #2313, opened 2022-07-07
Reduce course load time by loading modules in parallel

dimkard:


kantord:
nice!

kantord (changes requested):
@dimkard can you please run `poetry run black apps` and commit the changes? Looks like that's the only CI failure: https://kantord.semaphoreci.com/workflows/a294a9ab-56cc-4ecd-8ad0-7608c563417d?pipeline_id=7e57d5e5-d03f-490c-93a7-8391fa5a38b1

kantord:
it looks like the pytest test cases now never completely finish :thinking:

dimkard:
Not sure what happens, we can abandon this workaround and focus on the root cause #2314. :thinking:

kantord:
> Not sure what happens, we can abandon this workaround and focus on the root cause #2314. thinking

I will also try to look into this because I would say that it would be nice to use both solutions in combination in order to try to bring the loading times down to around 1s in order to be able to develop tools that work with the course data almost instantly

kantord:
closing this as seemingly there's no more work being done here. feel free to reopen if you restart the work

PR #2332, opened 2022-07-19
Adding two couple images

vandersdg:


kantord:
Hi, I like the images that you picked, but I'm curious why you picked 2 new images when we already have 3 for this word?

I created 3 images instead of 1 for each word so that it's not as boring and so that people don't memorize individual images, but I think 3 should be enough. I mean, each added image is going to grow the size of the course, and I have in mind creating the ability to use courses completely offline as well :thinking:

vandersdg:
@kantord Basically the couple2 page is not found, and there are no male+female couple images. In Portuguese the word couple only fits for man+woman or girl+boy, when the two are of the same gender we say "pair". But I agree that only 3 is enough, maybe a solution is to replace couple2 with one of my proposals.

kantord:
> @kantord Basically the couple2 page is not found, and there are no male+female couple images. In Portuguese the word couple only fits for man+woman or girl+boy, when the two are of the same gender we say "pair". But I agree that only 3 is enough, maybe a solution is to replace couple2 with one of my proposals.

ok, that's an interesting issue that I have not considered! btw, we don't directly link the pictures, they are actually cached in this repo to make sure we are not causing excess traffic for unsplash and that random 404 issues don't happen. 

this is how `couple2` actually shows up in the app: https://github.com/LibreLingo/LibreLingo/blob/main/apps/web/static/images/couple2.jpg

However, replacing the picture that now gives 404 could make sense! :ok_hand: 

By the way, do you think there's a reason specifically to upload a version (3 separate pictures) that are for Portuguese `casal`? :thinking:  and perhaps 3 more that are for Portuguese `par`? :thinking:

vandersdg:
Ohh if the image is cached it will work perfectly, I thought it didn't exist.

> By the way, do you think there's a reason specifically to upload a version (3 separate pictures) that are for Portuguese casal? thinking and perhaps 3 more that are for Portuguese par? 

Well, in my opinion, let's keep it simple :smile:. Couple is well illustrated by an image containing male+female; and pair is well illustrated by the other images (male+male and female+female). 
As the "couple2" image is working I will close this request.

PR #2457, opened 2022-09-28
Enable pylint duplicate-code rule

### LINKED ISSUE #1924: pylint config: enable "duplicate-code"

**How to solve**

**Why solve like this**

**Additional context**

jeremykenny:
## Description

Enabled the "duplicate-code" rule in the pylint config and changed the configuration settings for the rule. Refactored the following files to reduce the amount of duplicate code:

_apps/tools/tests/test_real_courses.py
apps/tools/tests/test_test_course.py_
Import the Settings object from fakes instead of declaring a new one

_apps/librelingo_audios/librelingo_audios/update_audios.py_
Pylint found lines 122-127 to be too similar to lines 73-78 of _conftest.py_. This seems to me to be an acceptable case of duplicate code, given that the lines of _conftest.py_ are used to test those from _update_audios.py_. However, the CI will not pass if they are left as is, so it was necessary to make some change. It seemed sensible to use a variable for TTS provider, given that more providers could be added in future.

fixes #1924

...
## Checklist

- [ ] The CI is passing

### DIFF
--- a/apps/tools/tests/test_real_courses.py
+++ b/apps/tools/tests/test_real_courses.py
@@ -1,5 +1,4 @@
 # pylint: skip-file
-import collections
 import os
 import pytest
 
@@ -19,13 +18,7 @@ from librelingo_types import (
     #    TextToSpeechSettings,
     #    Word,
 )
-
-Settings = collections.namedtuple(
-    "Settings",
-    [
-        "dry_run",
-    ],
-)
+from librelingo_fakes import fakes
 
 
 def load(name):
@@ -39,9 +32,7 @@ def load(name):
 
 def test_error_1(tmpdir):
     course = load("e1")
-    settings = Settings(
-        dry_run=False,
-    )
+    settings = fakes.settings_not_dry_run()
 
     with pytest.raises(RuntimeError) as err:
         export_course(str(tmpdir), course, settings)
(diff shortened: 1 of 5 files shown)


apps/tools/tests/test_real_courses.py
```diff
@@ -53,7 +46,7 @@ def test_error_1(tmpdir):
 
 def test_error_2(tmpdir):
     course = load("e2")
-    settings = Settings(
+    settings = fakes.Settings(
```
kantord, line 49: Hi @jeremykenny this PR is great :tada: but there's just one problem. Looks like this `fakes.Settings` is different, and does not have `dry_run`, and therefore these tests fail:

you can see that the CI fails because of this: https://kantord.semaphoreci.com/jobs/68410e8f-b9a1-4876-ae26-6d1a45df489d

you can also see the failures on your machine if you run the following command:

```bash
poetry run pytest
```

kantord: Please fix it in a way that the tests don't fail. If you get stuck, [I can jump on a call with you and help out](https://app.harmonizely.com/harmonizely-uoicqk/15min).

kantord (changes requested):

apps/tools/tests/test_real_courses.py
jeremykenny: Hi @kantord my apologies for the delay. I've figured it out, no call needed, but thank you for the support. I believe the CI will pass now.

kantord:
thanks @jeremykenny ! 

@allcontributors please add @jeremykenny for code

Issue #2479, opened 2022-10-10
GitHub Actions: run Python checks on OS X


kantord:
In [a previous pull request](https://github.com/LibreLingo/LibreLingo/pull/2465/commits/cc1d5809d0bcc61d0c12077be00751fefcf6708e), I ported our Python checks to GitHub Actions.

In that pull request, I only made the tests work on Ubuntu, not on OS X or Windows. This is because, while not much should change between the platforms, due to some small differences the pipeline was failing.

**In this task, please make sure that the CI passes with the OSX platform**.

To do that, you can extend the `matrix` property in this file:

https://github.com/LibreLingo/LibreLingo/blob/cc1d5809d0bcc61d0c12077be00751fefcf6708e/.github/workflows/python.yml#L5-L10

You can consult the GitHub Actions documentation to see how to add different operating systems to the matrix: https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#example-adding-configurations

**Now the hard part is, after making this change, you have to make sure that the CI passes, that is, you have to figure out the reason for the failure based on the errors and make changes so that everything passes**. 

If you need more context, you can always check out the [YouTube video](https://www.youtube.com/watch?v=JHjmguNqA1s&t=264s) where I write the code that you are changing now.

kantord:
for some reason I cannot assign you @vil02 :sweat_smile:

kantord:
Maybe it's only possible if you comment on the issue

vil02:
@kantord I think that this issue can be closed because of #2518

PR #2504, opened 2022-10-20
test: fix logic in `test_dry_run_doesnt_create_output_files`

vil02:
Brief description of the changes:
- the mock of `export_course` in the test [`test_dry_run_doesnt_create_output_files`](https://github.com/LibreLingo/LibreLingo/blob/0572d4ed7c263d41a902c0fd2ba868d0f25959da/apps/librelingo_json_export/tests/test_cli.py#L61),
- [`_export_skill`](https://github.com/LibreLingo/LibreLingo/blob/0572d4ed7c263d41a902c0fd2ba868d0f25959da/apps/librelingo_json_export/librelingo_json_export/export.py#L56) uses [`os.devnull`](https://docs.python.org/3/library/os.html#os.devnull),
- redundant calls to the constructor of `pathlib.Path` were removed.


kantord:
Hi @vil02 , can you please rebase this branch to the up to date `main`? There's a fix that should allow the rest of the CI pipelines to trigger.

vil02:
@kantord, rebasing is done.

apps/librelingo_json_export/librelingo_json_export/export.py
```diff
@@ -46,16 +47,18 @@ def _export_skill(export_path: str, skill: Skill, course: Course, settings=None)
             json.dump(skill_data, f, ensure_ascii=False, indent=2)
 
     if skill.introduction:
-        Path(Path(export_path) / "introduction").mkdir(parents=True, exist_ok=True)
+        (Path(export_path) / "introduction").mkdir(parents=True, exist_ok=True)
+        output_path = (
+            Path(os.devnull)
```
kantord, line 52: it seems like you are only doing this logic for the introduction files `.md` files but not for the json files in the output :thinking: which makes me think that for some reason the test is still wrong, because it doesn't fail. it should fail because I think according to the code, the JSON files are still created, so probably in the test JSON files would not be created (or perhaps, no files would be created) even if you don't pass the dry run flag.

Can you please try running the test, but change 

```python
invoke([*inputs, "--dry-run"])
```

to 

```python
invoke([*inputs])
```

This should definitely fail. If it doesn't then it means either the test setup is completely wrong, or the assertion is completely wrong.

So please also take a look at the assertion:

```python
assert os.listdir(".") == ["tmp"]
```

Perhaps `os.listdir(".")` will always return `["tmp"]` here no matter whether the files are generated or not.  Maybe it needs to be 

```python
assert os.listdir("./tmp") == [
  # something something
]
```

vil02: I already did this experiment before. The assertion was failing, because `os.listdir(".")` was returning something like `["tmp", "FakeValue(id=123)"]`.

Note that in case of `json`'s the function `json.dumps` is used (cf. [here](https://github.com/LibreLingo/LibreLingo/blob/61fc29c9c75b4240adbfba66b8a385c4f1420de3/apps/librelingo_json_export/librelingo_json_export/export.py#L44) and [here](https://github.com/LibreLingo/LibreLingo/blob/61fc29c9c75b4240adbfba66b8a385c4f1420de3/apps/librelingo_json_export/librelingo_json_export/export.py#L76)).

I will paste the logs of the _failed run_ in a moment.

kantord: I am thinking, though, that an entire file structure should be tested here, no? Perhaps more like an `os.walk()`, so as to make sure that actually no new files are created.

Or even do sth like 

```python
files_before_dry_run = set(os.walk())
# ... etc ...

assert set(os.walk()) == files_before_dry_run

```

vil02: In the newest change, I am comparing the result of `tuple(os.walk("."))`. I think this will also help in issue #2479, since "tmp" is not hard coded.

kantord: Gotcha, I'm just wondering if the test is working overall, does it fail if the json files are accidentally saved? :thinking: and it would be nice to refactor that part the same way for consistency, or even extract this logic somehow to make the code look neater. I can jump on a call with you to talk about it if you want: https://app.harmonizely.com/harmonizely-uoicqk

vil02: You are right, the test passes if some json files are accidentally created.

kantord:
> * should I use the [`os.devnull`](https://docs.python.org/3/library/os.html#os.devnull) _trick_ also for the `json` files instead of `json.dumps`?

yes, that would be great

> [this](https://github.com/LibreLingo/LibreLingo/blob/b3f645cab94b1fed533b1b0ecdc0739bf9e6a530/apps/librelingo_json_export/librelingo_json_export/cli.py#L38) seems to be redundant now (cf. the new function _ensure_output_dir in export.py).

I wonder if it breaks any tests if you remove it. Have you tried? 



> I think the that `test_dry_run_doesnt_create_output_files` is now doing the right thing,

fair enough, I mean, if you prefer I can merge this and if you have any further changes, we can merge it in a new PR

kantord:
Seems like there are some conflicts in `package.json`... I'm not sure how that's possible :sweat_smile:

vil02:
This has already become a _monster merge_. I think the best way would be to merge it now (I hope the there are no further conflicts) and I will make another PR with the remaining changes, i.e.
- using `os.devnull` trick for the json files,
- remove `ensure_output_directory` from `cli.py` (if possible),
- maybe add `test_no_dry_run_creates_output_files` (just in case...).

I think the conflict in `package.json` comes from my rebase about 2 days ago.

kantord:
Thank you @vil02 ! :heart:

Issue #2499, opened 2022-10-20
Code style: fix 2 instances of lines that are too long


kantord:
Earlier, pylint was set up to enforces that lines don't get too long. However, automatic formatting does not solve that issue in some cases, so 2 instances where this pylint rule is ignored were introduced. See this code for those 2 examples of `# pylint: disable=line-too-long` while also making sure that the tests no longer fail. So you need to manually edit the lines that create the problem.

https://github.com/LibreLingo/LibreLingo/blob/1da7e69de3bcc18c8a965f1cd480aa5394a94b54/apps/librelingo_yaml_loader/librelingo_yaml_loader/yaml_loader.py#L168-L214

rsk2:
@kantord, can I work on this?

kantord:
of course @rsk2 go for it! :rocket:

rsk2:
Hi @kantord, I have created a pull request for this issue.

PR #2512, opened 2022-10-24
Issue-2499: fix 2 instances of lines that are too long

rsk2:
To resolve issue #2499

### DIFF
--- a/apps/librelingo_yaml_loader/librelingo_yaml_loader/yaml_loader.py
+++ b/apps/librelingo_yaml_loader/librelingo_yaml_loader/yaml_loader.py
@@ -165,7 +165,6 @@ def _convert_word(raw_word) -> Word:
     )
 
 
-# pylint: disable=line-too-long
 def _convert_words(raw_words: List[Word]) -> List[Word]:
     """
     Converts each YAML word definition into Word() objects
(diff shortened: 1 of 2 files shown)


kantord:
Look good, thank you @rsk2 ! :tada: :100: :rocket:

kantord (changes requested):
Opps. I was wrong. It _looks_ good, but there's one problem: the tests are failing

https://github.com/LibreLingo/LibreLingo/actions/runs/3314578776/jobs/5481836901

Please fix the failing tests :pray: otherwise it cannot be merged

rsk2:
> Opps. I was wrong. It _looks_ good, but there's one problem: the tests are failing
> 
> https://github.com/LibreLingo/LibreLingo/actions/runs/3314578776/jobs/5481836901
> 
> Please fix the failing tests 🙏 otherwise it cannot be merged

@kantord those tests are failing because I formatted the expected output in order to reduce the line length . Any suggestion on how to fix this?

kantord:
@rsk2 looks like this could be the issue: https://stackoverflow.com/questions/28805567/line-continuation-wrapping-in-doctest it seems like doctest actually tests that the output strings are equivalent, that's why this is an error, because in the output, the whitespace is not there.

But it seems like it should be possible to instruct doctest to normalize whitespace, which would fix this problem. Looks like you can add this configuration option in order to fix it: https://docs.pytest.org/en/7.1.x/how-to/doctest.html#using-doctest-options

rsk2:
> @rsk2 looks like this could be the issue: https://stackoverflow.com/questions/28805567/line-continuation-wrapping-in-doctest it seems like doctest actually tests that the output strings are equivalent, that's why this is an error, because in the output, the whitespace is not there.
> 
> But it seems like it should be possible to instruct doctest to normalize whitespace, which would fix this problem. Looks like you can add this configuration option in order to fix it: https://docs.pytest.org/en/7.1.x/how-to/doctest.html#using-doctest-options

@kantord Thanks. I have made the suggested changes

kantord (approved):

kantord:
Good job @rsk2 thank you!!! :100:  it's merged! 

@allcontributors please add @rsk2 for code

PR #2518, opened 2022-10-26
chore: add macos-11

vil02:
I added `macos-11` to [python workflow](https://github.com/LibreLingo/LibreLingo/blob/a012439e84ba3ca2661ed4954fb33df79396da2f/.github/workflows/python.yml).

`poetry` is installed using [`snok/install-poetry@v1`](https://github.com/snok/install-poetry) - this will also work on windows. Installing `poetry` in this way, requires `bash` as a shell in every relevant step. I am open here for other suggestions.

This should solve #2479.

@ztjhz please have a look - it might be relevant to #2480.

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -7,7 +7,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: ["3.9"]
-        os: [ubuntu-18.04]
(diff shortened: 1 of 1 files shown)


kantord:
Thanks @vil02 !!!! :rocket:

PR #2520, opened 2022-10-26
Python ci windows check

### LINKED ISSUE #2480: GitHub Actions: run Python checks on Windows
In [a previous pull request](https://github.com/LibreLingo/LibreLingo/pull/2465/commits/cc1d5809d0bcc61d0c12077be00751fefcf6708e), I ported our Python checks to GitHub Actions.

In that pull request, I only made the tests work on Ubuntu, not on OS X or Windows. This is because, while not much should change between the platforms, due to some small differences the pipeline was failing.

**In this task, please make sure that the CI passes with the Windows platform**.

To do that, you can extend the `matrix` property in this file:

https://github.com/LibreLingo/LibreLingo/blob/cc1d5809d0bcc61d0c12077be00751fefcf6708e/.github/workflows/python.yml#L5-L10

You can consult the GitHub Actions documentation to see how to add different operating systems to the matrix: https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#example-adding-configurations

**Now the hard part is, after making this change, you have to make sure that the CI passes, that is, you have to figure out the reason for the failure based on the errors and make changes so that everything passes**. 

If you need more context, you can always check out the [YouTube video](https://www.youtube.com/watch?v=JHjmguNqA1s&t=264s) where I write the code that you are changing now.

ztjhz:
# Closes #2480

This pull request adds windows to the CI of GitHub Actions, while fixing issues prevalent on windows machine.

## Issue 1: Poetry is not added to `PATH` on windows

Solution: set the default shell to `bash` in `python.yml`
```yaml
defaults:
  run:
    shell: bash
```

## Issue 2: File separator is different on linux (`/`) and windows (`\`). Hence, hard-coded file paths will fail on windows.

Solution: utilise dynamic file paths
Example: change `"./output/challenges/hello.json"` to `str(Path("./output/challenges/hello.json"))`

## Issue 3: Regex string not escaped
For the regex matching in `pytest.raises(RuntimeError, match=expected_error)`, the `expected_error` was not escaped. This might cause errors in certain scenarios like windows file separator being `\\` . Hence, we need to escape it first by using `re.escape()`

## Issue 4: Character encoding not specified

Not specifying character encoding when reading files, might result in different special characters being read on windows and unicode decode errors.

Solution: specify the encoding:
```python
with open(path, encoding="utf-8") as yaml_file:
```

Please review the changes @kantord :)

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -7,7 +7,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: ["3.9"]
-        os: [ubuntu-18.04]
+        os: [ubuntu-18.04, windows-latest]
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v3
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -8,6 +8,9 @@ jobs:
       matrix:
         python-version: ["3.9"]
         os: [ubuntu-18.04, windows-latest]
+    defaults:
+      run:
+        shell: bash
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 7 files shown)


kantord:
Great job @ztjhz !!!!! :100: :rocket: :1st_place_medal:

kantord (approved):

PR #2522, opened 2022-10-26
refactor: simplify export.py

vil02:
I made a _cleanup_ of [`export.py`](https://github.com/LibreLingo/LibreLingo/blob/a012439e84ba3ca2661ed4954fb33df79396da2f/apps/librelingo_json_export/librelingo_json_export/export.py) and added relevant tests as discussed in PR #2504.

Things which I would like to drag your attention to:
- _invoke like functions_  in [`test_cli.py`](https://github.com/vil02/LibreLingo/blob/bf11561f5138ec786569e8f80d4e570ea29e8703/apps/librelingo_json_export/tests/test_cli.py) - do you see a way to treat `invoke`, `invoke_with_load_course_mock` and `invoke_with_no_mocks` in an uniform way?
- ~~is it OK to use `[]` as `settings` in [`test_export_prepare_output_path.py`](https://github.com/vil02/LibreLingo/blob/bf11561f5138ec786569e8f80d4e570ea29e8703/apps/librelingo_json_export/tests/test_export_prepare_output_path.py) while calling [`_prepare_output_path`](https://github.com/vil02/LibreLingo/blob/bf11561f5138ec786569e8f80d4e570ea29e8703/apps/librelingo_json_export/librelingo_json_export/export.py#L23)?~~
- the way how `no_dry_run_arg` is used in [`test_creates_excpected_files`](https://github.com/vil02/LibreLingo/blob/bf11561f5138ec786569e8f80d4e570ea29e8703/apps/librelingo_json_export/tests/test_cli.py#L102) a bit _hacky_ - does anything better come to your mind?


apps/librelingo_json_export/librelingo_json_export/export.py
```diff
@@ -20,9 +20,16 @@ def _ensure_output_dir(output_file_path):
...
 
-def _save_as_json_file(data, output_path):
-    _ensure_output_dir(output_path)
-    with open(output_path, "w", encoding="utf-8") as f:
+def _prepare_output_path(output_file_path, settings):
+    res = Path(os.devnull)
+    if not _is_dry_run(settings):
+        res = output_file_path
+        _ensure_output_dir(res)
+    return res
```
kantord, line 28:
```suggestion
def _prepare_output_path(output_file_path, settings):
    if settings.dry_run:
        return Path(os.devnull)

    _ensure_output_dir(output_file_path)
    return output_file_path
```

How about that? You can save one function and have an easier to read structure?

apps/librelingo_json_export/librelingo_json_export/export.py
```diff
@@ -20,9 +20,16 @@ def _ensure_output_dir(output_file_path):
...
+
+
+def _save_as_json_file(data, output_path, settings):
+    with open(_prepare_output_path(output_path, settings), "w", encoding="utf-8") as f:
         json.dump(data, f, ensure_ascii=False, indent=2)
```
kantord, line 33: What strikes me is that this function is specific to JSON files, but it does a logic that is not specific to JSON.

What if instead you just return a file pointer from this function, like so:

```suggestion
def _open_output_file(output_path, settings):
    return open(_prepare_output_path(output_path, settings), "w", encoding="utf-8")
```

Then you could use this for both JSON and YAML files, for example:

```python
    if skill.introduction:
        with _open_output_file(Path(export_path) / "introduction" / f"{slug}.md", settings) as f:
            f.write(skill.introduction)
```

so basically the API is very close to the builtin `open` but it handles the dry run and potentially other features

apps/librelingo_json_export/tests/test_cli.py
```diff
@@ -86,3 +96,18 @@ def test_dry_run_doesnt_create_output_files(
...
     invoke([*inputs, "--dry-run"])
     assert mocks["export_course"].call_count == 1
+
+
+@pytest.mark.parametrize("no_dry_run_arg", [["--no-dry-run"], []])
+def test_creates_excpected_files(fs, invoke_with_no_mocks, no_dry_run_arg):
+    input_path = Path(__file__).parent / "fixtures" / "fake_course"
+    fs.add_real_directory(input_path)
+    output_path = fakes.path()
+    assert not output_path.exists()
```
kantord, line 106: I think this assertion is not important or redundant, actually we don't really care if that folder exists, right? besides it should logically follow that it doesn't exist.

apps/librelingo_json_export/tests/test_cli.py
```diff
@@ -86,3 +96,18 @@ def test_dry_run_doesnt_create_output_files(
...
+    expected_files = {
+        output_path / _
+        for _ in ("courseData.json", "challenges/hello.json", "introduction/hello.md")
+    }
```
kantord, line 110:
```suggestion
    expected_files = {
        output_path / "courseData.json",
        output_path / "challenges/hello.json",
        output_path / "introduction/hello.md",
    }
```

No need to put some extra logic for 3 examples

apps/librelingo_json_export/librelingo_json_export/export.py
vil02: This is a real beauty!

apps/librelingo_json_export/tests/test_cli.py
vil02: Since `fs` is explicitly as one of the arguments, I agree.

apps/librelingo_json_export/tests/test_cli.py
vil02: The logic is not so complex here in my opinion. I would leave it as it is.

vil02:
I am not really sure why do the `python verifications` fail on windows.

apps/librelingo_json_export/tests/test_cli.py
kantord: it's true that the logic is not so complex but it doesn't read as easily, and does not seem to add any benefit, since we don't expect this list to grow or become dynamic

apps/librelingo_json_export/tests/test_export_is_dry_run.py
```diff
@@ -0,0 +1,12 @@
+import pytest
+from librelingo_json_export.export import _is_dry_run
+from librelingo_json_export.cli import Settings, DEFAULT_SETTINGS
+
+
+def test_returns_true_when_dry_run():
+    assert _is_dry_run(Settings(True))
```
kantord, line 7: ok, so in the end if we do get rid of this function (unless it's really needed for something else) then these tests won't be needed either

kantord: regarding this function I'm actually wondering why it's needed. How come `settings` can be `None`? I think it should always be supplied, no? 

I'm concerned that if we follow this logic, then a lot of similar fallback functions could be created if more settings value are used in a similar way where `settings` could be `None`.

If it's absolutely impossible to avoid having `settings` as `None`, then to prevent the proliferation of such functions that provide a fallback value for each settings value separately, why not create a logic that creates a fallback settings value?

```python
def get_settings(settings):
  if not settings:
    return DEFAULT_SETTINGS

  return settings
```

kantord:
btw looks like now there's a peculiar error in Windows, it seems related to `os.devnull` maybe this works on an actual windows machine, but not in github actions? In that case maybe there's already some solution for this? So far this is what I could find: https://github.com/ipython/ipython/issues/9023 but I know nothing about windows.

I'm just throwing ideas but maybe it's possible to monkey patch `os.devnull` with some actual fake folder using pytest :sweat_smile: I know it's quite a weird idea to *fake* /dev/null using real files but it could work for testing purposes

kantord:
> I am not really sure why do the `python verifications` fail on windows.

sorry I didn't see this comment before actually replying to it: https://github.com/LibreLingo/LibreLingo/pull/2522#issuecomment-1295313161

kantord (changes requested):
please address the remaining comments, otherwise looks good to me :+1:

apps/librelingo_json_export/tests/test_export_is_dry_run.py
vil02: Since `collections.namedtuple` is immutable, I decided to simply use `DEFAULT_SETTINGS` as a default arguments of `settings` in functions: `_export_course_skills`, `_export_skill`, `_export_course_data` and `export_course`. Before it was `None`. Such change does not change the _interface_ too much.

vil02:
@kantord what do you think about using [`io.StringIO`](https://docs.python.org/3/library/io.html#io.StringIO) instead of [`os.devnull`](https://docs.python.org/3/library/os.html#os.devnull)? (cf. vil02/LibreLingo#5)

kantord:
> @kantord what do you think about using [`io.StringIO`](https://docs.python.org/3/library/io.html#io.StringIO) instead of [`os.devnull`](https://docs.python.org/3/library/os.html#os.devnull)? (cf. [vil02#5](https://github.com/vil02/LibreLingo/pull/5))

it's perfect, let's do it that way. you have to merge that thing in your fork to have it in this PR, right?

apps/librelingo_json_export/tests/test_export_is_dry_run.py
kantord: amazing, that's the best solution

vil02:
> > @kantord what do you think about using [`io.StringIO`](https://docs.python.org/3/library/io.html#io.StringIO) instead of [`os.devnull`](https://docs.python.org/3/library/os.html#os.devnull)? (cf. [vil02#5](https://github.com/vil02/LibreLingo/pull/5))
> 
> it's perfect, let's do it that way. you have to merge that thing in your fork to have it in this PR, right?

Yes and it is done.

kantord:
Nice! Thanks @vil02 !!!!

PR #2524, opened 2022-10-27
Ensure that YAML loader works without PyYAML C bindings

dimkard:
By making use of CSafeLoader we require that the LibYAML C bindings of PyYAML are installed. If these bindings are not present, this commit ensures that the YAML Loader still works. However, performance will not be as good as with the C bindings.

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -7,7 +7,10 @@ jobs:
       fail-fast: false
       matrix:
         python-version: ["3.9"]
-        os: [ubuntu-18.04, macos-11]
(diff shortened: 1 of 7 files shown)


kantord:
Hi @dimkard it seems like there are some conflicts as well as some failures in the CI

dimkard:
Well, pylint does not like this line `apps/librelingo_yaml_loader/librelingo_yaml_loader/yaml_loader.py:46:0: C0301: Line too long (190/100) (line-too-long)`

I run `poetry run black apps/librelingo_yaml_loader/` but nothing changes.

I have no insight on how to make pylint and mypy happy anymore ...

PR #2525, opened 2022-10-27
Ensure that YAML loader works without PyYAML C bindings

dimkard:
By making use of CSafeLoader we require that the LibYAML C bindings of PyYAML are installed. If these bindings are not present, this commit ensures that the YAML Loader still works. However, performance will not be as good as with the C bindings.

### DIFF
--- a/apps/librelingo_yaml_loader/librelingo_yaml_loader/yaml_loader.py
+++ b/apps/librelingo_yaml_loader/librelingo_yaml_loader/yaml_loader.py
@@ -1,3 +1,4 @@
+import logging
 import collections
 import os
 import re
@@ -26,6 +27,8 @@ from yaml.constructor import SafeConstructor
 
 from ._spelling import _convert_hunspell_settings, _run_skill_spellcheck
 
+logger = logging.getLogger("librelingo_yaml_loader")
+
 
 def add_bool(self, node):
     return self.construct_scalar(node)
@@ -42,7 +45,7 @@ def _load_yaml(path: Path):
 
             return load(yaml_file, Loader=CSafeLoader)
         except ImportError:
-            print(
+            logger.warning(
                 """Warning! PyYAML LibYAML C bindings are not installed.
              Course loading still works, but it will be slower.
              For more details, check https://github.com/yaml/pyyaml#Installation"""
--- a/apps/librelingo_yaml_loader/librelingo_yaml_loader/yaml_loader.py
+++ b/apps/librelingo_yaml_loader/librelingo_yaml_loader/yaml_loader.py
@@ -50,7 +50,7 @@ SafeConstructor.add_constructor("tag:yaml.org,2002:bool", add_bool)
 
 def _load_yaml(path: Path):
     """Helper function for reading a YAML file"""
-    with open(path) as yaml_file:
+    with open(path, encoding="utf-8") as yaml_file:
         return load(yaml_file, Loader=SafeLoader)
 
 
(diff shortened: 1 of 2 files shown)


kantord:
Seems like this specific change is not compatible with Windows, in the CI failure I can see:

```
   def decode(self, input, final=False):
>       return codecs.charmap_decode(input,self.errors,decoding_table)[0]
E       UnicodeDecodeError: 'charmap' codec can't decode byte 0x81 in position 400: character maps to <undefined>

```

apps/librelingo_yaml_loader/librelingo_yaml_loader/yaml_loader.py
```diff
@@ -22,10 +22,6 @@
 )
 from yaml import load
 
-try:
-    from yaml import CSafeLoader as SafeLoader
-except ImportError:
```
kantord, line 27: how come this part didn't already work as a fallback for `CSafeLoader`?

apps/librelingo_yaml_loader/librelingo_yaml_loader/yaml_loader.py
```diff
@@ -40,8 +36,20 @@ def add_bool(self, node):
...
 def _load_yaml(path: Path):
     """Helper function for reading a YAML file"""
-    with open(path, encoding="utf-8") as yaml_file:
-        return load(yaml_file, Loader=SafeLoader)
+    with open(path) as yaml_file:
+        try:
+            from yaml import CSafeLoader
+
+            return load(yaml_file, Loader=CSafeLoader)
+        except ImportError:
```
kantord, line 44: How come this logic works here, but it didn't work when it was on the top of the file? :thinking:

apps/librelingo_yaml_loader/librelingo_yaml_loader/yaml_loader.py
```diff
@@ -40,8 +36,20 @@ def add_bool(self, node):
...
+            print(
```
kantord, line 45: Well, it seems like this is printing it to standard output, which is not the ideal place to show this warning in, as warnings/debug messages are not part of the intended output

kantord: perhaps you should use the `logging` module for this? https://stackoverflow.com/a/6579522

kantord (changes requested):

apps/librelingo_yaml_loader/librelingo_yaml_loader/yaml_loader.py
dimkard: The first approach I tried is this one:
```
try:
    from yaml import CSafeLoader as SafeLoader
except ImportError:                                                               
    from yaml import SafeLoader                                                   
```
and it results in mypy error:
```
error: Incompatible import of "SafeLoader" (imported name has type "Type[SafeLoader]", local name has type "Type[CSafeLoader]")
```

apps/librelingo_yaml_loader/librelingo_yaml_loader/yaml_loader.py
dimkard: TBH I ended up with this not so elegant approach to avoid mypy errors.

If it is OK to add explicit `# type: ignore` on the import statements, I can go back to the original top-of-the-file approach.

kantord: Well, if this has working types then it's better

apps/librelingo_yaml_loader/librelingo_yaml_loader/yaml_loader.py
dimkard: yes, just the mypy error.

Issue #2438, opened 2022-09-25
Add JSON schema validation for courses


kantord:

djgoreli:
Hi I would be interested in looking at this task!

kantord:
partially solved in https://github.com/LibreLingo/LibreLingo/pull/2788

PR #2612, opened 2022-12-12
Add initial schema validation for modules

djgoreli:
Added the jsonschema package to the project and initial schema validation logic to the yaml_loader for validating module .yaml files. Also created a schemas directory within apps/librelingo_yaml_loader and added a JSON schema to validate the format of module .yaml files.

Contributes to #2438, but there is more to be done adding the loading for courses and skills.

I had a few uncertainties with implementing these changes as follows:
- Is it alright to have schemas located in apps/librelingo_yaml_loader and is there any way to get around hard coding this path to load the schema json files from? I considered it being within the courses directory but I don't think there is a way for it to be there without duplicating the same schema file across multiple courses
- There is additional error checking in the _load_modules function I edited that I believe will become redundant with the schema validation, but I was not sure if I should remove it yet

### DIFF
--- a/apps/librelingo_yaml_loader/librelingo_yaml_loader/yaml_loader.py
+++ b/apps/librelingo_yaml_loader/librelingo_yaml_loader/yaml_loader.py
@@ -383,17 +383,15 @@ def _load_module(path: str, course: Course):
     filepath = Path(path) / "module.yaml"
     data = _load_yaml(filepath)
 
-    # Hardcoded is not great, should path lead to some other place?
     schemapath = "../../apps/librelingo_yaml_loader/schemas/module.json"
     with open(schemapath) as fp:
         schema = json.load(fp)
 
-    # Should I take out unnecessary exception handling below?
     try:
         jsonschema.validate(data, schema)
     except jsonschema.ValidationError as validation_error:
         raise RuntimeError(
-            f'There is an error with the schema at the following file path: {filepath}'
+            f"There is an error with the schema at the following file path: {filepath}"
         ) from validation_error
 
     try:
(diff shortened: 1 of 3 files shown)


apps/librelingo_yaml_loader/librelingo_yaml_loader/yaml_loader.py
```diff
@@ -380,6 +382,18 @@ def _load_module(path: str, course: Course):
     """
     filepath = Path(path) / "module.yaml"
     data = _load_yaml(filepath)
+
+    schemapath = "../../apps/librelingo_yaml_loader/schemas/module.json"
```
kantord, line 386: > Is it alright to have schemas located in apps/librelingo_yaml_loader and is there any way to get around hard coding this path to load the schema json files from? I considered it being within the courses directory but I don't think there is a way for it to be there without duplicating the same schema file across multiple courses

regarding this, Python has a built in feature for this and Poetry makes sure that these things are properly set up in the final package, so this solution should work:

https://docs.python.org/3/library/pkgutil.html#pkgutil.get_data

or you could use `pkg_resources` for this

apps/librelingo_yaml_loader/librelingo_yaml_loader/yaml_loader.py
```diff
@@ -380,6 +382,18 @@ def _load_module(path: str, course: Course):
...
+    with open(schemapath) as fp:
+        schema = json.load(fp)
+
+    try:
+        jsonschema.validate(data, schema)
+    except jsonschema.ValidationError as validation_error:
+        raise RuntimeError(
+            f"There is an error with the schema at the following file path: {filepath}"
```
kantord, line 394: will this hide the particular details of the validation error? isn't it better to simply let the whole process fail at this point? (so simply removing the try-except?)

kantord:
Look good to me, but it would be nice to have at least an integration test for this, or something

PR #2627, opened 2022-12-31
Add metadata to courses configuration file

dimkard:
As discussed in #2538, add metadata:
    - source, e.g., "English"
    - target, e.g., "Spanish"
    - name, e.g., "spanish-from-english"
    - description, e.g., "Spanish for English speakers"

### DIFF
--- a/config/courses.json
+++ b/config/courses.json
@@ -8,7 +8,11 @@
     },
     "deploy": true,
     "devtoolsEnabled": true,
-    "inProduction": true
+    "inProduction": true,
+    "name": "spanish-from-english",
+    "source": "English",
+    "target": "Spanish",
+    "description": "Spanish for English speakers"
   },
   {
     "url": "https://github.com/szabgab/LibreLingo-Judeo-Spanish-from-English/archive/refs/heads/main.zip",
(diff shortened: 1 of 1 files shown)


config/courses.json
```diff
@@ -8,7 +8,11 @@
     },
     "deploy": true,
     "devtoolsEnabled": true,
-    "inProduction": true
+    "inProduction": true,
+    "name": "spanish-from-english",
+    "source": "English",
```
kantord, line 13: I would use something more verbose but more specific:

```suggestion
    "sourceLanguage": {
      "name": "English",
    },
```

config/courses.json
```diff
@@ -8,7 +8,11 @@
...
+    "target": "Spanish",
+    "description": "Spanish for English speakers"
```
kantord, line 15: this could be automatically generated, right?

dimkard: TBH I did not know that courses.json is auto generated. IIUC, it is [generate.py](https://github.com/LibreLingo/LibreLingo/blob/main/apps/tools/librelingo_tools/generate.py) that creates it, right? 

So, this PR does not make sense, and we need a new one for the script? And just use what's available in [course](https://github.com/LibreLingo/LibreLingo/blob/d46672e5f91fa0b484c8131dae7bc823e96d21cf/apps/tools/librelingo_tools/generate.py#L85)?

kantord: no, I don't mean courses.json is auto-generated. I mean, that contents of that field could be auto-generated (in some code or library based on this file) rather than being specified manually here.

Actually it is something that should probably be translated even :thinking:

kantord: The PR makes sense overall

kantord: I changed my mind and I will approve and merge this. The reason is that the description field could be used as a generic comment field rather than an automatically generated title for the course

PR #2822, opened 2023-04-17
updated dependencies

rahulbollisetty:
this fixes [Issue 2682](https://github.com/LibreLingo/LibreLingo/issues/2682)

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -2,14 +2,14 @@
 
 [[package]]
 name = "astroid"
-version = "2.15.2"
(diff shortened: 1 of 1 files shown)


kantord:
hi @rahulbollisetty this indeed works, however it seems like it doesn't actually change the version requirement in pyproject.toml, it will just update the lockfile to show what version those requirements will resolve to

actually, as per the poetry documentation, the proper way to update the package is to use the following command: 

```bash
poetry add mypy@latest --dev
```

PR #2827, opened 2023-04-20
add-share-to-mastodon-button

KevinReardon:
Implemented share to Mastodon button and added to the fanfare page.

### DIFF
--- a/apps/web/src/components/FanfareScreen.svelte
+++ b/apps/web/src/components/FanfareScreen.svelte
@@ -12,7 +12,7 @@
   import Column from "lluis/Column.svelte"
   import Columns from "lluis/Columns.svelte"
   import Title from "lluis/Title.svelte"
-
+  import MastodonButton from "./MastodonButton.svelte"
   export let courseURL
   export let skillId
   export let stats
@@ -56,6 +56,7 @@
             <div class="bottom">
               <Title size={5}>Excited about LibreLingo?</Title>
               <TwitterButton />
+              <MastodonButton/>
             </div>
           </div>
         </Column>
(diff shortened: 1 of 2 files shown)


kantord:
Hi @KevinReardon good job! 

Looks good, just 2 small problems, could you check them?

First, the mastodon icon does not seem to load:

2023-04-20-092325_504x304_scrot

I think this is because that icon is not loaded in the app, but [it does exist in the icon package that we use ](https://fontawesome.com/icons/mastodon?f=brands&s=solid)so it should be easy to fix. 

I think that you can just change **src/client.js** in the web app package. Right now it has something like this:

```typescript
import * as sapper from "@sapper/app"
import "./mystyles.scss"
import "./i18n"
import { library, dom } from "@fortawesome/fontawesome-svg-core"
import {
  faVolumeUp,
  faCheckSquare,
  faDumbbell,
  faStar,
  faUser,
  faLock,
  faEnvelope,
  faHeart,
  faSpinner,
  faCircleArrowRight,
  faCircleArrowLeft,
} from "@fortawesome/free-solid-svg-icons"
import { faTwitter } from "@fortawesome/free-brands-svg-icons"

library.add(faVolumeUp)
library.add(faCheckSquare)
library.add(faDumbbell)
library.add(faTwitter)
library.add(faStar)
library.add(faUser)
library.add(faLock)
library.add(faEnvelope)
library.add(faHeart)
library.add(faSpinner)
library.add(faCircleArrowRight)
library.add(faCircleArrowLeft)
dom.watch()
```

I think you can just import the mastodon icon there and then it will work.

The second problem, might not really be a problem: when I click the button I am lead to a page where the form is not prefilled at all. I understand that the url of the mastodon instance cannot be prefilled, but I'm wondering, can at least some message be prefilled? For example "Hey, I practiced this and that skill on LibreLingo!" or sth like that?

2023-04-20-092729_1047x629_scrot

PR #2831, opened 2023-04-21
pylint config: enable "invalid-name"

PeteLast:
I went through and corrected all but 2 errors on the invalid-name Pylint. The two that remain are related to changing 2 file names to snake-case, but I believe that changes critical access information, so I chose to wait on those.

I tested my implementation manually and locally, and did not write any new tests. I also commented out the invalid-name portion of the Pylint ignore file and did not delete it, so it could be referenced easily.

This is related to and fixes "Pylint config: enable "invalid-name" #1928"

### DIFF
--- a/apps/librelingo_json_export/tests/test_cli.py
+++ b/apps/librelingo_json_export/tests/test_cli.py
@@ -25,12 +25,12 @@ def mocks(mocker):
 
 
 @pytest.fixture
-def invoke(mocks, fs):
-    def f(args):
+def invoke(mocks):
+    def file(args):
         runner = CliRunner()
         return runner.invoke(main, args, catch_exceptions=False)
 
-    return f
+    return file
 
 
 @pytest.fixture
@@ -39,21 +39,21 @@ def load_course_mock(mocker):
 
 
 @pytest.fixture
-def invoke_with_load_course_mock(load_course_mock, fs):
-    def f(args):
+def invoke_with_load_course_mock(load_course_mock):
+    def file(args):
         runner = CliRunner()
         return runner.invoke(main, args, catch_exceptions=False)
 
-    return f
+    return file
 
 
 @pytest.fixture
-def invoke_with_no_mocks(fs):
-    def f(args):
+def invoke_with_no_mocks():
+    def file(args):
         runner = CliRunner()
         return runner.invoke(main, args, catch_exceptions=False)
 
-    return f
+    return file
 
 
 def test_yaml_to_json_loads_correct_course(mocks, inputs, invoke):
@@ -72,7 +72,7 @@ def test_yaml_to_json_has_help_text(mocks, inputs, invoke):
     assert main.help
 
 
-def test_creates_output_directory_if_it_doesnt_exist(mocks, inputs, invoke, fs):
+def test_creates_output_directory_if_it_doesnt_exist(mocks, inputs, invoke):
     output_path = "foo/500/bar"
     invoke([inputs[0], output_path])
     assert os.path.isdir(output_path)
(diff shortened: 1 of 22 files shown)


apps/librelingo_json_export/tests/test_cli.py
```diff
@@ -101,10 +101,10 @@ def test_dry_run_calls_real_export_course(mocks, inputs, invoke):
 
 
 @pytest.mark.parametrize("no_dry_run_arg", [["--no-dry-run"], []])
-def test_creates_excpected_files(fs, invoke_with_no_mocks, no_dry_run_arg):
+def test_creates_excpected_files(f_s, invoke_with_no_mocks, no_dry_run_arg):
     input_path = Path(__file__).parent / "fixtures" / "fake_course"
-    fs.add_real_directory(input_path)
-    fs.add_real_directory(os.path.dirname(jsonschema.__file__))
+    f_s.add_real_directory(input_path)
```
kantord, line 106: regarding this, I think it should be added to `"good-names"` like this: https://stackoverflow.com/a/37799332

my reason for that is because that is the name of a mock that I think wasn't defined in our repo, right? so I think we can allow that, especially if it's related to a convention

apps/librelingo_json_export/tests/test_integration.py
```diff
@@ -10,15 +10,15 @@
 
 
 def read_json_file(path):
-    with open(path) as fp:
-        return json.load(fp)
+    with open(path) as f_p:
```
kantord, line 13: elsewhere `f_p` was instead named `file` which I think makes more sense,  I think you could name all of them `file` or `file_pointer` :thinking:

apps/librelingo_utils/librelingo_utils/utils.py
```diff
@@ -44,10 +44,10 @@ def clean_word(word: Word):
     """
     Remove punctuation and other special characters from a word.
     """
-    MATCH_NON_WORD_CHARACTERS_BEGINNING = regex.compile("^[^\\w']+")
-    MATCH_NON_WORD_CHARACTERS_END = regex.compile("[^\\w']+$")
-    return MATCH_NON_WORD_CHARACTERS_BEGINNING.sub(
-        "", MATCH_NON_WORD_CHARACTERS_END.sub("", word)
+    match_non_word_characters_beginning = regex.compile("^[^\\w']+")
```
kantord, line 47: could these, instead of being renamed, simply be extracted outside the function? My understanding is that all caps snake case is a convention for constants. I guess, technically these are constants but I imagine they are not recognized as such because they are in a function scope, not in the module scope. So I think if you just move this out of the function as a "global" const, it should work without renaming: `MATCH_NON_WORD_CHARACTERS_BEGINNING = regex.compile("^[^\\w']+")`

apps/tools/librelingo_tools/lili.py
```diff
@@ -135,8 +135,8 @@ def export_main_html_page(course, count, reldir, html_dir):
         course=course,
         repository_url=get_repository_url(course),
     )
-    with open(os.path.join(html_dir, "modules.html"), "w") as fh:
-        fh.write(html)
+    with open(os.path.join(html_dir, "modules.html"), "w") as f_h:
```
kantord, line 138: this is the same as the case that I mentioned above, I I think this is just a file pointer, it could be named with either a custom name that is appropriate for this use case, or a generic name like you did above (`file` or `file_pointer`)

apps/web/package.json
```diff
@@ -12,7 +12,7 @@
     "jest": "jest src",
     "types": "sapper build && tsc && svelte-check --treshold error",
     "installCourse": "./scripts/installExternalCourse.sh",
-    "installAllExternalCourses": "python ./scripts/installExternalCourses.py ../../config/courses.json",
+    "installAllExternalCourses": "python3 ./scripts/installExternalCourses.py ../../config/courses.json",
```
kantord, line 15: I think this is not correct, I think that Python 2 is considered legacy software at this point and it should be called `python2` while `python3` should be called `python`. 

However I think by replacing it with `poetry run python` I think it will be even managed through poetry, so then I think it should work even if it's an outdated OS that still considers Python2 to be the default version

kantord (changes requested):

kantord:
> The two that remain are related to changing 2 file names to snake-case, but I believe that changes critical access information, so I chose to wait on those.

can you show me what those are?

PR #2944, opened 2023-07-01
Voice samples for card and selection challenges, fixes #617

### LINKED ISSUE #617: Voice samples for "Which of these is" exercices
**Is your feature request related to a problem? Please describe.**

Duolinguo can play a voice sample when the user clicks on a card in the `Which of these is` exercice.
I find it useful for languages where the pronunciation is hard to get.

**Describe the solution you'd like**

In addition to images, we should be able to add voice samples for each options.

**Describe alternatives you've considered**

Displaying the information using the phonetic alphabet

killuaDev:
Voice samples now play whenever the selected card or phrase changes to a non-null value, audio is added to these challenge types as well.

I haven't edited the feature files because I'm not sure exactly what details I would put in there, something like this maybe?
```gherkin 
And I click a card
And I hear the word
```

### DIFF
--- a/apps/librelingo_json_export/librelingo_json_export/challenge_types.py
+++ b/apps/librelingo_json_export/librelingo_json_export/challenge_types.py
@@ -56,7 +56,7 @@ def get_short_input_challenge(source: Word, course: Course):
     ]
 
 
-def get_cards_challenge(word: Word, _):
+def get_cards_challenge(word: Word, course: Course):
     return [
         {
             "type": "cards",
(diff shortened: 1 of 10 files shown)


kantord:
> I haven't edited the feature files because I'm not sure exactly what details I would put in there, something like this maybe?
> 
> ```
> And I click a card
> And I hear the word
> ```

yes, this would work! :rocket:  you can handle it in a second merge request, if you want

kantord (changes requested):
it looks good to me! :rocket: 

however, there are some failures in the tests. it looks like it's because a lot of test cases test the format of the json, and the audios need to be included.

could you please fix it? and let me know if you need help, we can jump on a call: https://app.harmonizely.com/harmonizely-uoicqk

killuaDev:
Yeah I'll try to have a look at those soon, I've been a bit busy lately but hopefully I'll have some time

PR #2965, opened 2023-07-12
feat(yaml-loader): load modules concurrently using threads

photonbit:
Resurrected PR from #2313

The error on that PR was coming from opening file descriptors in a multiprocessing environment that were not available afterwards for the original process. By using threads we can make the loading concurrent, but I am not sure if this will actually improve loading time.


kantord:
@allcontributors please add @photonbit for code

PR #3175, opened 2023-11-21
Fix invalid json in librelingo_yaml_loader

sublime-coleslaw:
scripts/verifyTestCourseJsonFiles.sh breaks on invalid json

### DIFF
--- a/apps/librelingo_yaml_loader/librelingo_yaml_loader/skill.json
+++ b/apps/librelingo_yaml_loader/librelingo_yaml_loader/skill.json
@@ -64,7 +64,7 @@
         },
         "required": [
           "Word",
-          "Translation",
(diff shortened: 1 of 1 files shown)


kantord:
wow, so this means that this script is not being run in the CI :open_mouth:  I think that this script is meant to be executed in the CI to make sure that breaking changes cannot be accidentally merged

sublime-coleslaw:
the deployment and python workflows run the verifyTestCourseJsonFiles script, but a different issue (#3176) has been breaking CI before the verifyTestCourseJsonFiles script even runs, so I think that's why this comma wasn't caught. #3177 adds a json linter in pre-commit

kantord:
thanks @sublime-coleslaw !

Issue #3159, opened 2023-11-08
Add .pre-commit-config.yaml and use pre-commit to run the various linters, formatters, and checks


yelircaasi:
[pre-commit](https://pre-commit.com/) is a tremendous project that has greatly improved my workflow - easy to install and easy to use. I would be happy to implement it here and open a pull request, I just wanted to see first if that is something people would be on board with. Its benefits are twofold - firstly, it ensures code standards are met before each commit. Secondly, it provides an easy command to run all relevant linters etc. at any point during the development process.

kantord:
> [pre-commit](https://pre-commit.com/) is a tremendous project that has greatly improved my workflow - easy to install and easy to use. I would be happy to implement it here and open a pull request, I just wanted to see first if that is something people would be on board with. Its benefits are twofold - firstly, it ensures code standards are met before each commit. Secondly, it provides an easy command to run all relevant linters etc. at any point during the development process.

yes, this would be amazing. I am also already using pre-commit in some other project and I can attest that it is amazing, so a PR is definitely welcome! 

Also I think in the config file, I think the linting/formatting tools could be replace with https://github.com/astral-sh/ruff

PR #3177, opened 2023-11-22
Use pre-commit for linting

sublime-coleslaw:
#3159 

This adds a dev dependency for pre-commit, and a basic pre-commit configuration for linting with ruff. The second commit fixes the various issues that pre-commit complained about.

Tests still aren't run using pre-commit, and none of the github actions have been switched to use pre-commit. I'm not sure if that would be a good idea since pre-commit usually isn't used to run tests.


kantord (changes requested):
Looks good to me, but this file needs to be changed as well to make the pipeline work: https://github.com/LibreLingo/LibreLingo/blob/main/.github/workflows/python.yml

sublime-coleslaw:
I added ruff as a poetry dependency and changed the workflow to use ruff instead of black/pylint

kantord:
hi @sublime-coleslaw it seems like it is still failing? https://github.com/LibreLingo/LibreLingo/actions/runs/6958213826/job/19029305201?pr=3177

are you having trouble running these checks in your fork? maybe I could help you with that

sublime-coleslaw:
whoops! forgot to check that command locally before pushing. unfortunately github doesn't seem to want to run any of the workflows in my fork on push, and i can't create a PR in my fork. if you know of a workaround, i'd appreciate help getting it to run :)

kantord:
> github doesn't seem to want to run any of the workflows in my fork on push, and i can't create a PR in my fork. if you know of a workaround, i'd appreciate help getting it to run :)

If I understand correctly, it should be enough for you to [enable Github Actions in your fork](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository), and then it should run the github action using your fork. But that probably does not apply to the "pull request" trigger, only to the "push" trigger I guess. (I'm not sure about it)

So I think this might actually be my fault, as it seems the `push` trigger is not enabled in the repo. I think

PR #3276, opened 2024-02-05
Added UI translation for Serbian language

novta:


novta:
What is problem with those tests? There are no free space in POedit group, I expected simple json translation would be enough.
How to translate UI to Serbian and deploy?

kantord:
> What is problem with those tests? There are no free space in POedit group, I expected simple json translation would be enough. How to translate UI to Serbian and deploy?

what do you mean there is no free space in the POedit group, you mean you are not able to join and contribute the translation keys there?

Adding the JSON is an option, I mean, it is possible to manually import it into POeditor, but the correct way should be to contribute through POeditor and automatically export the translation keys into the repo.

If it's not possible to contribute through POeditro that is a problem, it means we need to find an alternative to poeditor

kantord:
The failing tests are not a problem with this PR or Poeditor, so don't worry about those

novta:
When I try to join to community https://poeditor.com/join/project/jKiKfen2Ye with my account I get next message `This project can't accept new contributors at this time. Please try again later.`
image

novta:
So if test are not consequence of localization it can be merged. Is that correct?

EmanuelLoos:
The reason why it does not work through POeditor is that there are some translation keys missing. Therefore POeditor says that the translation is complete when it actually is not. This PR seems to add some more translation keys where the English strings are currently hardcoded. I think it should be merged.

PR #3281, opened 2024-02-08
Update courses.json

RandomPenguin1778:
Added Hindi course

### DIFF
--- a/config/courses.json
+++ b/config/courses.json
@@ -238,5 +238,20 @@
     "source": "English",
     "target": "Dutch",
     "description": "Dutch for English speakers"
+  },
+  {
+    "url": "https://github.com/LibreLingo/LibreLingo-HI-from-EN/archive/refs/tags/v.1.0.0.zip",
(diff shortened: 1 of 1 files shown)


config/courses.json
```diff
@@ -238,5 +238,20 @@
     "source": "English",
     "target": "Dutch",
     "description": "Dutch for English speakers"
+  },
+  {
+    "url": "https://github.com/LibreLingo/LibreLingo-HI-from-EN/archive/refs/tags/v.1.0.0.zip",
```
kantord, line 243: Hi, before I merge this PR, be aware that this will deploy the v.1.0.0 tag, not the latest passing version. So it means that any new update to the course will require a new PR. If this is intended I can merge it, but this should probably be a temporary measure and over the long term the updates should be deployed automatically

RandomPenguin1778: hmm. how about I translate and make the course first, and then open a seperate pull request?

Issue #3278, opened 2024-02-06
Rusyn language for Serbian speakers


novta:
Hello,

This is a great project and idea, I would like to participate in it.

I am translating the spanish-from-english version of the course to rusyn-from-serbian, 
The Cyrillic alphabet is used for both languages so I guess it will not be a problem.

Should I simply add course reference when it is ready to the list of courses?

novta:
There is the repo https://github.com/novta/LibreLingo-RSK-from-SR with course

PR #3288, opened 2024-02-11
 Rusyn language for Serbian speakers #3278 

novta:


kantord:
at the moment it cannot be merged because of a rewrite effort, but once the rewrite effort is at a stage where there is actually a translation implemented I will be happy to merge it (probably the same code will kinda work after rebase)

Issue #1143, opened 2021-03-03
Add challenges for characters and their pronunciation


adnan360:
**Is your feature request related to a problem? Please describe.**

Sometimes seeing letters on screens is not enough to learn how to write them on paper. Writing is an important part of learning a language and without such guidelines, learning is not complete. Right now there's no way to show how to write a letter.

Actually the [idea](https://github.com/kantord/LibreLingo/pull/1121#issuecomment-786842864) came from @kantor and I'm just opening an issue for it:
> I don't know the Bengali script, so there's no way for me to know how to pronounce or read the words that I see. LibreLingo should add some way to teach the characters in Bengali script, and how to combine those characters together to form words/sentences.

And possibly there should be challenges to test what they've learned.

**Describe the solution you'd like**
@kantor describes it better:
> ...One of the main idea that I have for solving this is adding another section along with Words and Phrases. It would be something like Characters. And it would generate challenges similar to the ones for words, but it would be only about characters and their pronunciation

Previous quote talks about "teaching" and this one about "challenges". So there's 2 aspects to it.

I have prepared a [letter handwriting animation](https://github.com/kantord/LibreLingo/pull/1121#issuecomment-789793975) code to aid with the teaching part. Maybe it can help in the process or not (feel free to discuss).

**Describe alternatives you've considered**
I've thought about video recording a handwriting gesture. But it can't be done by everyone or won't be of same quality and it would need more bandwidth to serve.

**Additional context**
Well, since we're talking about challenges here, how would we collect a correct answer? Through drawing? If so, we'd need OCR or something like that to validate input. Or is there other simple ways? Feel free to discuss possibilities.

kantord:
> Well, since we're talking about challenges here, how would we collect a correct answer? Through drawing? If so, we'd need OCR or something like that to validate input. Or is there other simple ways? Feel free to discuss possibilities.

As a first, super-simple idea I can imagine simply a version of the listening challenges. You just have to listen to an audio and select out of a list, which character you heard. In the case of Bengali, maybe this could be a combination of 2 characters, so you'd basically hear a syllable and choose with character combination it belongs to.

The reverse of that could also be possible, basically seeing a character and choosing which sound matches it.

kantord:
I'm gonna pin this, it should not be terribly hard to solve so maybe someone is interested in implementing it

pur80a:
As I imagine it, a minimal character `skill.yaml` could look like this: 
```
Skill:
[...]

PR #3405, opened 2024-10-17
Add Characters to LibreLingo python scripts

Skivling:
This is the preliminary work for #1143, implementing them in Python. Later, once @kantord is up to it in the web app rewrite they can be implemented there.

I have made some assumptions for what best to name the titles in Yaml and functions, which I am fine changing if there is a better option.

I have also added a property in Characters for their International Phonetic Alphabet (IPA) pronounciations, as I believe this would be useful to have below or above the characters for some learners, especially those who are very into linguistics and/or learning a conlang in a LibreLingo course. 
The IPA value is not required in the schema so not including it won't cause any errors.

I hope unicode and such doesn't cause any bugs or errors such as a new character not displaying properly. 🤞

Currently Character()s in a skill look like this:
```yaml
New Characters:
   
   - Character: ā
     Transliteration: aa
     IPA: /aː/
   
   - Character: Г
     Transliteration: H
     IPA: /ɦ/
```

And here they are in Python:
```python
    # A long sound macron on an 'a' in Te Reo Māori
    >>> Character("ā", "aa", "/aː/")
    Character(character='ā', transliteration='aa', ipa_pronounciation='/aː/')

    # The Г character in Ukranian (which uses the Cyrillic script).
    >>> Character("Г", "H", "/ɦ/")
    Character(character='Г', transliteration='H', ipa_pronounciation='/ɦ/')
```

I have yet to finish the in-code documentation of the functions, then adding all this to the proper docs, hence this PR is a draft. I also haven't tested anything yet.

### DIFF
--- a/docs/courses/course.md
+++ b/docs/courses/course.md
@@ -112,7 +112,7 @@ Modules:
   - `License > Name`: Full license name under which your course is made available. In most cases, it's ok to keep it as is.
   - `License > Short name`: Short name for the license, e.g., `CC BY-SA 4.0`
   - `License > Link`: URL to reach the full text of the license, e.g., `https://creativecommons.org/licenses/by-sa/4.0/legalcode`
-- `Special characters`: An array of special characters that might not be present on a typical English keyboard.
+- `Special characters`: An array of special characters that might not be present on a typical English keyboard. Note that this is separate from `New Characters` included inside some skills.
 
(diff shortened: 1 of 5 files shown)


Skivling:
I have made the IPA field a list because it's quite common for a character to represent a few distinct but similar sounds. Same as the transliteration field.

Should the transliteration field be renamed as Transliterations (plural) to indicate this?

